PR #1, opened 2023-03-15
feat(operator): Add new dragonfly k8s operator

Pothulapati:
This commit adds a new Kubernetes operator written in Golang into the repo, that can create and manage DragonFly Instances for Users. Operator is a tool to manage the lifecycle of DragonFly Instances in the cluster. It allows for easy creation, upgrade of dragonfly instances without having to understand the underlying details.

This commit of the operator only focuses on the installation part. More configurations, fail-over part will be added later.

\## Running the Operator locally

The Operator can be ran locally by running the following commands.

```bash
kind create cluster

cd operator

make install

make run

kubectl apply -f config/samples/v1alpha1_dragonflydb.yaml
```

You should see the Operator create the relevant
statefulset and service for the Dragonfly Instance.

```bash
kubectl get statefulsets
```

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -2,7 +2,3 @@ resources:
 - manager.yaml
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-images:
-- name: controller
-  newName: ghcr.io/pothulapati/flydragon/operator
-  newTag: latest


internal/controller/dragonfly_controller.go
```diff
@@ -0,0 +1,114 @@
...
+		resources, err := resources.GetDragonflyResources(ctx, &df)
+		if err != nil {
+			log.Error(err, "could not get resources")
+			return ctrl.Result{}, err
+		}
+
+		// create all resources
+		for _, resource := range resources {
+			if err := r.Create(ctx, resource); err != nil {
+				log.Error(err, "could not create resource")
```
ashotland: Is there something helpful in the resource object to add to the log line?

internal/controller/util.go
```diff
@@ -0,0 +1,158 @@
...
+			if err := markReplica(ctx, c, pod, masterIp); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func markReplica(ctx context.Context, client client.Client, pod corev1.Pod, masterIp string) error {
```
ashotland: What do you think about 
func (pod corev1.Pod) configureAsReplica(...)
Then we can write pod.configureAsReplica(.... masterIp)

In any case consider renaming to configureAsReplica

internal/controller/util.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	if statefulSet.Status.ReadyReplicas == *statefulSet.Spec.Replicas {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func findHealthyAndMarkActive(ctx context.Context, c client.Client, db *resourcesv1.Dragonfly) error {
```
ashotland: I'd call it configureReplication

And (similar to another comment below) I like the syntax 
func (*resourcesv1.Dragonfly dragonfly) (....)
So we can write dragonfly.configureReplication(....

Any any case lets not call the dragonfly parameter db :)

internal/controller/util.go
Pothulapati: `corev1.Pod` is a non-local type so we can't define methods on those :/ I'm actually thinking to have some abstraction with the fail-over part as these are used more! We will have this discussion in that PR.

PR #3, opened 2023-03-17
feat(ci): Add tests into CI

Pothulapati:
part of #8 

This PR adds `make test` into the CI build pipeline. This
does not add any tests recently

### DIFF
--- a/internal/controller/dragonfly_controller_test.go
+++ b/internal/controller/dragonfly_controller_test.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	dragonflydbiov1alpha1 "github.com/dragonflydb/dragonfly-operator/api/v1alpha1"
+	resourcesv1 "github.com/dragonflydb/dragonfly-operator/api/v1alpha1"
 	"github.com/dragonflydb/dragonfly-operator/internal/resources"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


internal/controller/dragonfly_controller_test.go
```diff
@@ -0,0 +1,124 @@
...
+			}, &svc)
+			Expect(err).To(BeNil())
+
+			// Wait until Dragonfly object is marked initialized
+			waitForDFInitialised(ctx, k8sClient, name, namespace, 2*time.Minute)
+			waitForStatefulSetReady(ctx, k8sClient, name, namespace, 2*time.Minute)
+
+			// Check if there are relevant pods with expected roles
+			var pods corev1.PodList
+			err = k8sClient.List(ctx, &pods, client.InNamespace("default"), client.MatchingLabels{
```
Pothulapati: We are only testing for resource creation here, but we should also verify roles which I'm thinking can be in a separate PR

internal/controller/dragonfly_controller_test.go
```diff
@@ -0,0 +1,124 @@
...
+				resources.KubernetesPartOfLabelKey: "dragonfly",
+			})
+			Expect(err).To(BeNil())
+
+			Expect(pods.Items).To(HaveLen(4))
+		})
+	})
+})
+
+func waitForDFInitialised(ctx context.Context, c client.Client, name, namespace string, maxDuration time.Duration) error {
```
ashotland: It's Dragonfly so Df or Dfly
But Dragonfly is best :)

Issue #8, opened 2023-03-17
Automated Testing in the Operator


Pothulapati:
It is important to have automated testing for the Operator so
that we can be sure about the changes. These tests
should include both unit and e2e tests testing
majority of the functionality

PR #14, opened 2023-03-21
feat(health): Add tests to verify fail-over

Pothulapati:
Part of #8 

This Commit adds `health_controller_test.go` that essentially
tests for automatic fail-over of master if the pod goes down.
It also verifies the same for replica's as pods are restarted.

### DIFF
--- a/internal/controller/replication.go
+++ b/internal/controller/replication.go
@@ -19,7 +19,9 @@ type inClusterClient struct {
 	client client.Client
 }
 
-func NewInClusterClient(client client.Client) *inClusterClient {
+// NewReplicationClient returns a new replication client
+// that works with in the cluster and is the default.
+func NewReplicationClient(client client.Client) *inClusterClient {
 	return &inClusterClient{
 		client: client,
 	}
--- a/internal/controller/suite_test.go
+++ b/internal/controller/suite_test.go
@@ -91,7 +91,7 @@ var _ = BeforeSuite(func() {
 	err = (&HealthReconciler{
 		Client:            k8sManager.GetClient(),
 		Scheme:            k8sManager.GetScheme(),
-		ReplicationClient: newOutClusterClient(clientset, *testEnv.Config),
+		ReplicationClient: newOutClusterReplicationClient(clientset, *testEnv.Config),
 	}).SetupWithManager(k8sManager)
 	Expect(err).ToNot(HaveOccurred())
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -110,7 +110,7 @@ func main() {
 	if err = (&controller.HealthReconciler{
 		Client:            mgr.GetClient(),
 		Scheme:            mgr.GetScheme(),
-		ReplicationClient: controller.NewInClusterClient(mgr.GetClient()),
+		ReplicationClient: controller.NewReplicationClient(mgr.GetClient()),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "Health")
 		os.Exit(1)
--- a/internal/controller/dragonfly_instance.go
+++ b/internal/controller/dragonfly_instance.go
@@ -58,7 +58,7 @@ func GetDragonflyInstanceFromPod(ctx context.Context, c client.Client, pod *core
 
 	// Use InClusterConfigurer by default
 	if rClient == nil {
-		rClient = NewInClusterClient(c)
+		rClient = NewReplicationClient(c)
 	}
 
 	return &DragonflyInstance{
--- a/internal/controller/testutil.go
+++ b/internal/controller/testutil.go
@@ -23,7 +23,7 @@ type outClusterClient struct {
 	config    rest.Config
 }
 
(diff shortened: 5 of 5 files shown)


internal/controller/suite_test.go
```diff
@@ -76,18 +78,22 @@ var _ = BeforeSuite(func() {
...
-		err = (&HealthReconciler{
-			Client: k8sManager.GetClient(),
-			Scheme: k8sManager.GetScheme(),
-		}).SetupWithManager(k8sManager)
-		Expect(err).ToNot(HaveOccurred())
-	*/
+	err = (&HealthReconciler{
+		Client:     k8sManager.GetClient(),
+		Scheme:     k8sManager.GetScheme(),
+		Configurer: newOutclusterMarker(clientset, *testEnv.Config),
```
Pothulapati: `ReplicationMarker` is necessary abstraction as this Operator runs outside the cluster during tests which means it can't access the pod IP directly (to mark pods as master or redis). This interface allows to have a new `outClusterMarker` that uses the Kubeconfig and performs a port-forward to access the pod IP.

internal/controller/replication.go
```diff
@@ -0,0 +1,74 @@
...
+	"context"
+	"fmt"
+
+	"github.com/dragonflydb/dragonfly-operator/internal/resources"
+	"github.com/go-redis/redis"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type replicationMarker interface {
```
ashotland: ReplicationClient ?

internal/controller/replication.go
```diff
@@ -0,0 +1,74 @@
...
+	replicaOf(ctx context.Context, pod *corev1.Pod, masterIp string) error
+	replicaOfNoOne(ctx context.Context, pod *corev1.Pod) error
+}
+
+type inClusterMarker struct {
+	client client.Client
```
ashotland, line 19: I don't think we must have this wrapper

NewInclusterConfigurer (which I'd rename to newReplicationClient) can return *client.Client
And replicaOf , replicatOfNoOne can extend it 

Right?

internal/controller/health_controller.go
```diff
@@ -34,6 +34,8 @@ import (
 type HealthReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
+
+	Configurer replicationMarker
```
ashotland: Let try to be as minimalistic as possible with terminology 

ReplicationClient replicationClient  ?

internal/controller/suite_test.go
ashotland: Isn't it possible to communicate with all the df server through the headless service which will overcome this issue more elegantly ?

Pothulapati: Service is still not accessible outside the cluster here as we are not exposing it in any way (i.e ingress or something). Even if we do, We still don't get pod-level access i.e the service allows us to access only the master but the Replication actions have to be performed across all pods i.e not just master. WDYT?

ashotland: I believe headless service provides access to each individual pod, but your point about out of cluster accessibility is valid.

Pothulapati: > I believe headless service provides access to each individual pod

Hmm. Yep you get the direct pod ip with no extra cluster IP (& load balancing) but you are not in control of choosing which pod you want to access. We need that here right? as we want to run specific commands on specific pods. 🤔

internal/controller/replication.go
Pothulapati: Sounds right!

internal/controller/suite_test.go
ashotland: I thought it does expose a dns name per pod, in any case, as you say, using a service by itself won't solve the out of cluster accessibility issue.

internal/controller/replication.go
Pothulapati: We can't extend `client.Client` to have new methods as its a non-local type 🤔

internal/controller/replication.go
```diff
@@ -0,0 +1,74 @@
...
+type replicationClient interface {
+	replicaOf(ctx context.Context, pod *corev1.Pod, masterIp string) error
+	replicaOfNoOne(ctx context.Context, pod *corev1.Pod) error
+}
+
+type inClusterClient struct {
+	client client.Client
+}
+
+func NewInclusterClient(client client.Client) *inClusterClient {
```
ashotland: Fix CamelCasing 

Maybe also just call this one  NewReplicaitonClient.

Pothulapati: We have two different client's right? The `Incluster` which is the default as well as `OutCluster` in `testuti.go` that we will use for tests. Without this naming difference, It could be confusing? 

(Fixing the camel casing)

ashotland: Mean that the factory function for the default one can be just called NewReplicationClient and in the test NewOutClusterReplicationClient.

Also consider adding the word Replicaiton to the factory functions as it is less clear in the context of the other files that uses it.

PR #15, opened 2023-03-22
feat(Operator): Handle updates to resources

### LINKED ISSUE #4: Support updates to `Dragonfly` resource
Currently, Only the initial installation is supported and verified. The 
same should also be done for the resource fields updates

Pothulapati:
Fixes #4 

This commit adds the logic to handle updates to the
`Dragonfly` resource.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -125,7 +125,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err = (&controller.DFPodLifeCycleReconciler{
+	if err = (&controller.DfPodLifeCycleReconciler{
 		Client:        mgr.GetClient(),
 		Scheme:        mgr.GetScheme(),
 		EventRecorder: eventRecorder,
--- a/internal/controller/dragonfly_pod_lifecycle_controller.go
+++ b/internal/controller/dragonfly_pod_lifecycle_controller.go
@@ -32,7 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-type DFPodLifeCycleReconciler struct {
+type DfPodLifeCycleReconciler struct {
 	client.Client
 	Scheme        *runtime.Scheme
 	EventRecorder record.EventRecorder
@@ -47,7 +47,7 @@ type DFPodLifeCycleReconciler struct {
 // This reconcile events focuses on configuring the given pods either as a `master`
 // or `replica` as they go through their lifecycle. This also focus on the failing
 // over to replica's part to make sure one `master` is always available.
-func (r *DFPodLifeCycleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *DfPodLifeCycleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := log.FromContext(ctx)
 
 	log.Info("Received", "pod", req.NamespacedName)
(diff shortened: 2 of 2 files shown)


internal/controller/dragonfly_controller.go
```diff
@@ -89,6 +84,42 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		log.Info("updating existing resources")
+		newResources, err := resources.GetDragonflyResources(ctx, &df)
+		if err != nil {
+			log.Error(err, "could not get resources")
+			return ctrl.Result{}, err
+		}
+
+		// update all resources
+		for _, resource := range newResources {
+			if err := r.Update(ctx, resource); err != nil {
```
ashotland, line 112: Does this maintain HA? 
I.e. what happens if the master node happens to be removed? 
Does it avoid unnecessary changes to pod, e.g if I add/remove a replica all existing stay in tact?

internal/controller/dragonfly_instance.go
```diff
@@ -64,8 +64,19 @@ func GetDragonflyInstanceFromPod(ctx context.Context, c client.Client, pod *core
...
+		Name:      d.df.Name,
+		Namespace: d.df.Namespace,
+	}, d.df); err != nil {
+		return "", err
+	}
+
+	return d.df.Status.Phase, nil
+}
+
 func (d *DragonflyInstance) configureReplication(ctx context.Context) error {
```
ashotland: Nit: rename d to df or dfi or dfcr as d can be read as dictionary

internal/controller/dragonfly_instance.go
```diff
@@ -64,8 +64,19 @@ func GetDragonflyInstanceFromPod(ctx context.Context, c client.Client, pod *core
...
-	d.log.Info("configuring replication")
+	d.log.Info("Configuring replication")
```
ashotland: I got a bit confused here as we have PhaseConfiguringReplication but if IIUC this can be called also in phase ready.
Shouldn't we actually check that Status.Phase != PhaseConfiguringReplication to avoid parallel/inter-lived executions of this function
and then set Status.Phase = PhaseConfiguringReplication ?
Should we lock on when changing status.phase?

internal/controller/health_controller.go
```diff
@@ -77,6 +77,18 @@ func (r *HealthReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 
+	if df.df.Status.Phase == PhaseConfiguringReplication {
+		log.Info("Dragonfly object is already configuring replication")
+		log.Info("Something went wrong. reconfigure")
```
ashotland: hmm so there's an implied invariant the reconcile shouldn't happen with Status.Phase == PhaseConfiguringReplication why?

Maybe we should formalise this better.

internal/controller/health_controller.go
```diff
@@ -86,20 +98,18 @@ func (r *HealthReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	role, ok := pod.Labels[resources.Role]
 	// New pod with No resources.Role
 	if !ok {
-		log.Info("Replication is not configured yet")
+		log.Info("Replication is not configured yet", "phase", df.df.Status.Phase)
```
ashotland: "No replication role was set yet, ..."

internal/controller/health_controller.go
```diff
@@ -86,20 +98,18 @@ func (r *HealthReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 				log.Error(err, "could not initialize replication")
 				return ctrl.Result{}, err
 			}
 
 			r.EventRecorder.Event(df.df, corev1.EventTypeNormal, "Replication", "configured replication for first time")
-
-		} else if df.df.Status.Phase == PhaseReady {
-			// Probably a pod restart do the needful
+		} else {
+			// Pod event either from a restart or a resource update (i.e less/more replicas)
```
ashotland, line 100: Does this imply Status.Phase == Ready?
can we assert on that or something?

cmd/main.go
```diff
@@ -125,7 +125,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err = (&controller.HealthReconciler{
+	if err = (&controller.DFPodLifeCycleReconciler{
```
ashotland: Nit: Acronym casing should be DfP

PR #20, opened 2023-03-31
fix(tests): Run the operator separately

Pothulapati:
Currently, Our operator test suite ran the operator locally
which meant we had to set port-forwards to talk to induvidual
pods when required to be marked as replicas or as masters. These
port-forwards are error-prone and induced a lot of retry logic
into the application code making errors confusing.

This commit fixes that by essentially updating the suite_test.go
to run the tests inside a already operator installed cluster.

### DIFF
--- a/internal/controller/health_controller_test.go
+++ b/internal/controller/health_controller_test.go
@@ -19,26 +19,16 @@ package controller
 import (
 	"context"
 	"fmt"
-	"io"
-	"net/http"
-	"os"
-	"strings"
 	"time"
 
 	dragonflydbiov1alpha1 "github.com/dragonflydb/dragonfly-operator/api/v1alpha1"
 	"github.com/dragonflydb/dragonfly-operator/internal/resources"
-	"github.com/go-redis/redis"
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


internal/controller/health_controller_test.go
```diff
@@ -93,15 +89,11 @@ var _ = Describe("Health Reconciler", func() {
 			Expect(pods.Items).To(HaveLen(replicas + 1))
 
 			// Get the pods along with their roles
-			for i, pod := range pods.Items {
+			for _, pod := range pods.Items {
 				role, ok := pod.Labels[resources.Role]
 				// error if there is no label
 				Expect(ok).To(BeTrue())
-				// verify the role to match the label
-				redisRole, err := getRole(ctx, clientset, cfg, &pod, 6381+i)
```
Pothulapati, line 101: We have stopped checking for the redis role inside the pod as it requires port-forwarding which is prone to failures in CI. There are multiple issues around this i.e https://github.com/kubernetes-sigs/kind/issues/2953, and are usually flaky requiring restarts.

Just checking for the role label should be enough, as the same functions that run the command on the pods also add the label only if running the command passes. So, Just checking for label should be enoug.

PR #23, opened 2023-04-03
feat(operator): bump dragonfly version to `v1.1.1`

Pothulapati:
`v1.1.1` involves various fixes around replication, etc that
we want to  use and its a official release.

### DIFF
--- a/internal/resources/version.go
+++ b/internal/resources/version.go
@@ -17,5 +17,5 @@ limitations under the License.
 package resources
 
 const (
-	Version = "v1.0.0"
+	Version = "v1.1.1"
 )


internal/resources/version.go
```diff
@@ -17,5 +17,5 @@ limitations under the License.
 package resources
 
 const (
-	Version = "v1.0.0"
+	Version = "v1.1.1"
```
ashotland, line 20: I think this should go into the spec.

Pothulapati: The spec has the `spec.image` field right? through which you can not only change the version but also the registry/image if needed? This is just the default value that we pick if no `spec.image` field is set.

ashotland: 👍

PR #29, opened 2023-04-03
fix(dragonfly): follow kubernetes replicas convention

### LINKED ISSUE #26: follow replicas primitive in Kubernetes
Our replica is different from replicas in Kubernetes environment. We should use the same kubernetes semantics instead

Pothulapati:
Fixes #26 

Currently, In Kubernetes replicas involve both master and read-only instances.
Up until now, DF replica's meant only the read only and excluded master.
This is confusing for k8s users and hence being changed as the operator
focuses on k8s users.


internal/resources/resources.go
```diff
@@ -69,7 +66,7 @@ func GetDragonflyResources(ctx context.Context, db *resourcesv1.Dragonfly) ([]cl
 			},
 		},
 		Spec: appsv1.StatefulSetSpec{
-			Replicas:    &replicas,
+			Replicas:    &db.Spec.Replicas,
```
ashotland: not urgent this pr but we agreed not to use the term db.

PR #33, opened 2023-04-04
fix(df_instance): error out when can't find a master pod

Pothulapati:
When the master pod is not found, the controller should error out
instead of moving forward and causing other errors.


internal/controller/dragonfly_instance.go
```diff
@@ -101,6 +101,11 @@ func (dfi *DragonflyInstance) configureReplication(ctx context.Context) error {
 		}
 	}
 
+	if master == "" {
+		dfi.log.Info("Couldn't find a master pod")
```
ashotland: "Couldn't find a healthy pod to configure as master"

PR #34, opened 2023-04-04
fix(df_pod_controler): Requeue when errors and remove `configuring-replication` phase

Pothulapati:
This commit has the following changes that improve the reliability
of the dragonfly_pod_lifecycle_controller:
- *Requeue when errors with commands*: As we know commands are prone
  to timeout, error when running commands. This commit adds a
  requeue when errors with commands. So, that we retry configuring
  replication.
- *Remove `configuring-replication` phase*: This phase is not
  necessary. We can just use `resources-created` and `ready` phases
  so that we have a 0, or 1 state and act accordingly.

### DIFF
--- a/internal/resources/const.go
+++ b/internal/resources/const.go
@@ -50,6 +50,8 @@ const (
 	// KubernetesPartOfLabel is the name of a higher level application this one is part of
 	KubernetesPartOfLabelKey = "app.kubernetes.io/part-of"
 
+	MasterIp string = "master-ip"
+
 	Role string = "role"
 
 	Master string = "master"


internal/controller/dragonfly_pod_lifecycle_controller.go
```diff
@@ -107,47 +106,53 @@ func (r *DfPodLifeCycleReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			log.Info("master is being removed. re-configuring replication")
+			if err := dfi.configureReplication(ctx); err != nil {
+				log.Error(err, "couldn't find healthy and mark active")
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, err
 			}
+			r.EventRecorder.Event(dfi.df, corev1.EventTypeNormal, "Replication", "Updated master instance")
+		} else if pod.Labels[resources.Role] == resources.Replica {
+			log.Info("replica is being deleted. nothing to do")
+		}
+	} else {
```
Pothulapati, line 146: This case is essentially a "catch-all" case. `checkAndConfigureReplication` function checks for all cases i.e master, replica pods exist, replicas are configured to the right new master, and confirms. Requeues if there is a error so we retry. Even after adding this, sometimes we do fall into https://github.com/dragonflydb/dragonfly/issues/1032 but that's a df issue.

internal/controller/dragonfly_controller.go
```diff
@@ -57,14 +56,13 @@ type DragonflyReconciler struct {
...
 	log := log.FromContext(ctx)
-
 	var df dfv1alpha1.Dragonfly
 	if err := r.Get(ctx, req.NamespacedName, &df); err != nil {
 		log.Info(fmt.Sprintf("could not get the Dragonfly object: %s", req.NamespacedName))
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	log.Info("Reconciling Dragonfly object")
+	log.Info("Reconciling Dragonfly object V1")
```
ashotland: Remove V1 - just added it to validate that I am running this version of the operator.

internal/controller/dragonfly_instance.go
```diff
@@ -190,37 +188,103 @@ func (dfi *DragonflyInstance) configureReplica(ctx context.Context, pod *corev1.
...
+			redisRole = strings.Trim(strings.Split(line, ":")[1], "\r")
+		}
 	}
 
+	if redisRole != resources.Replica {
+		return false, nil
+	}
+
+	var redisMasterIp string
+	// check if the pod is a replica, check if it is connected to the right master
```
ashotland: If the pod is a replica (no check)

internal/controller/dragonfly_instance.go
```diff
@@ -190,37 +188,103 @@ func (dfi *DragonflyInstance) configureReplica(ctx context.Context, pod *corev1.
...
+			return err
+		}
+	}
+
+	if len(podRoles[resources.Replica]) != len(pods.Items)-1 {
+		dfi.log.Info("incorrect number of replicas", "replicas", podRoles[resources.Replica])
+
+		// configure non replica pods as replicas
+		for _, pod := range pods.Items {
+			if pod.Labels[resources.Role] != resources.Replica {
```
ashotland: Isn't master also in pods.Items? if so aren't we overriding it?

internal/controller/dragonfly_pod_lifecycle_controller.go
```diff
@@ -107,47 +106,53 @@ func (r *DfPodLifeCycleReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		// is something wrong? check if all pods have a matching role and revamp accordingly
+		log.Info("Role exists already. checking if something is wrong", "pod", fmt.Sprintf("%s/%s", pod.Namespace, pod.Name), "role", role)
```
ashotland: Got a non deletion event for a pod with an existing role...

internal/controller/dragonfly_pod_lifecycle_controller.go
ashotland: checkAndConfigureReplication and configureReplicaiton seems to overlap quit a bit no?
I wonder if they can be merged elegantly, please think about it a bit, but if it is not simple we can refactor later.

internal/resources/version.go
```diff
@@ -17,5 +17,5 @@ limitations under the License.
 package resources
 
 const (
-	Version = "v1.1.1"
+	Version = "alpha-ubuntu"
```
ashotland: change?
maybe we don't need this at all?
must we supply a default image or can we require specifying it in the spec?

internal/controller/dragonfly_instance.go
Pothulapati: you are right!

internal/controller/dragonfly_instance.go
Pothulapati: that was a wrong comment. We only call `checkReplicaRole` if its a replica here. removing this comment.

internal/controller/dragonfly_pod_lifecycle_controller.go
Pothulapati: I would prefer to refactor later. Yes, They are similar but we use `configureReplication` specifically when we understand the trigger i.e master is not present but `checkAndConfigureReplication` is a catch-all case for all cases where we get a pod trigger with role and no deletion where we aren't sure what's wrong.

internal/resources/version.go
Pothulapati: You are right. removing this debug commit

PR #45, opened 2023-04-12
fix(readme): increase demo resource requirements

Pothulapati:
Resource requirements too low causing OOMKills sometimes


ashotland (commented):
LGTM

PR #49, opened 2023-04-17
fix(dfi): check pod ip before marking

Pothulapati:
Now that we mark pods as they get instanced without waiting for all
resources to be created the code cannot expect all pods to be ready
while it configures replication. This adds an extra pod IP Check.

If there is a pod without IP, the controller will wait for the next
pod event which gets triggered automatically once Kubernetes adds the
IP.

### DIFF
--- a/internal/controller/dragonfly_instance.go
+++ b/internal/controller/dragonfly_instance.go
@@ -270,8 +270,7 @@ func (dfi *DragonflyInstance) checkAndConfigureReplication(ctx context.Context) 
 
 		// configure non replica pods as replicas
 		for _, pod := range pods.Items {
-			if pod.Labels[resources.Role] != resources.Replica && pod.Labels[resources.Role] != resources.Master {
-				dfi.log.Info("configuring pod as replica", "pod", pod.Name)
+			if pod.Labels[resources.Role] == "" {
 				if err := dfi.configureReplica(ctx, &pod); err != nil {
 					return err
 				}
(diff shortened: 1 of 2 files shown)


internal/controller/dragonfly_instance.go
```diff
@@ -87,7 +87,7 @@ func (dfi *DragonflyInstance) configureReplication(ctx context.Context) error {
 	var master string
 	var masterIp string
 	for _, pod := range pods.Items {
-		if pod.Status.Phase == corev1.PodRunning && pod.Status.ContainerStatuses[0].Ready && pod.Labels[resources.Role] != resources.Master && pod.DeletionTimestamp == nil {
+		if pod.Status.Phase == corev1.PodRunning && pod.Status.ContainerStatuses[0].Ready && pod.Labels[resources.Role] != resources.Master && pod.DeletionTimestamp == nil && pod.Status.PodIP != "" {
```
ashotland: Can we have Status.ContainerStatuses[0].Ready while the IP was not set?

If so doesn't it mean we need to change our readiness probe?

Pothulapati: Our readiness (& healthcheck) probes are both the same and check if the PING commands in the container work. Pod IP assignment happens from K8s and can happen before or after. I've seen some cases where the IP was empty and we were trying to connect, and hence the additional check.

ashotland (approved):

PR #52, opened 2023-04-17
fix(dragonfly): check pod status before marking replica

Pothulapati:
This adds the same checks that we use everywhere else before
marking a replica as ready. This should prevent the replica
from being marked as ready before the pod is actually ready.


Pothulapati:
added this change into https://github.com/dragonflydb/dragonfly-operator/pull/49

PR #55, opened 2023-04-18
fix(readme): Update manifests to use `v0.0.3`

Pothulapati:
Updates the manifests used in readme demos to use the
`v0.0.3` version of the Operator.


Pothulapati:
This should be merged only after the release is published.

ashotland (approved):

Issue #48, opened 2023-04-16
rollout: wait until replication


Pothulapati:
Rollout should make sure replication is ready before it moves to the next pod. 
Should start from master.

ashotland:
A couple of hopeful helpful notes:
* I think that we should wait until replication reaches "stable state" (full sync is complete) - @adiholden / @dranikpg is this exposed somehow with any command?
* We should rely on requeue for the polling and not introduce a polling loop inside the the reconcile function (probable means expending the state)

dranikpg:
> is this exposed somehow with any command?

Not that I know one. There is `WAIT` in Redis, but its for the all replicas.

In replication tests, we just send write traffic as long as we get an 'instance is in loading state' error back. But we shouldn't do that with the operator, we can invent a new command for this.

> suld make sure replication is ready before it moves to the next pod.

What's crucial about it? DF support multiple full sync stages in parallel.

ashotland:
@dranikpg 
> What's crucial about it? DF support multiple full sync stages in parallel.
In a rolling update we want to avoid getting to updating the master too quickly before replicas are in a some sensible state. Makes sense?

Pothulapati:
To give the context from Kubernetes side, As we create a underlying statefulset for the Dragonfly instance We follow the same method that is used whenever a statefulset is updated i.e The update starts with the highest numbered pod and goes until the last one. There is no way for the statefulset to know which pod is master and do something different about it.

What I would propose is, As rolling out a pod is essentially terminating the current one, and creating a new one (with latest configuration) in its place, We should make sure a Dragonfly pod termination happens gracefully i.e If its a master, It should make sure Replicas are upto date before it can terminate fully. We could also increase the time period here to allow for it to complete fully.

@ashotland Are there any reasons why you think we should start from Replicas and then move to master? I'm not yet sure if we can choose the order of the rollout with Kubernetes. If we start with the master, The lifecycle controller recognises that the current master is going down, and selects a new one automatically. So, There should not be any downtime.

dranikpg:
>  i.e If its a master, It should make sure Replicas are upto date before it can terminate fully

Redis has `WAIT` for this and shutdown by default waits for replicas to sync - we don't. I honestly don't know how you achieve consistency reliably without it 😅

> In a rolling update we want to avoid getting to updating the master too quickly before replicas are in a some sensible state. Makes sense?

It does. Nonetheless making the sensible state less sensible sounds also reliable.

[...]

PR #59, opened 2023-04-23
fix(rollout): Update rollout strategy to be `OnDelete`

Pothulapati:
Fixes https://github.com/dragonflydb/dragonfly-operator/issues/48

This commit updates the statefulset's upgrade strategy to be `OnDelete` which means that the statefulset will not be automatically upgraded when the pod template changes. Instead, The Operator will delete each pod in the statefulset one at a time. This gives us the following benefits:
- The order of pod rollout is controlled by the Operator which means we can start from replicas and go until master.
- The Operator can wait for the pod to be ready and also be in stable sync before it moves to the next pod.

### DIFF
--- a/e2e/dragonfly_controller_test.go
+++ b/e2e/dragonfly_controller_test.go
@@ -351,6 +351,14 @@ func isDragonflyInphase(ctx context.Context, c client.Client, name, namespace, p
 		return false, nil
 	}
 
+	// Ready means we also want rolling update to be false
+	if phase == controller.PhaseReady {
+		// check for replicas
+		if df.Status.IsRollingUpdate {
+			return false, nil
+		}
+	}
+
 	if df.Status.Phase == phase {
 		return true, nil
 	}
--- a/e2e/dragonfly_controller_test.go
+++ b/e2e/dragonfly_controller_test.go
@@ -215,7 +215,7 @@ var _ = Describe("Dragonfly Reconciler", Ordered, func() {
 		})
 
 		It("Update to image should be propagated successfully", func() {
-			newImage := resources.DragonflyImage + ":v0.17.0"
+			newImage := resources.DragonflyImage + ":v1.1.0"
 			// Update df to the latest
 			err := k8sClient.Get(ctx, types.NamespacedName{
 				Name:      name,
--- a/internal/controller/dragonfly_controller.go
+++ b/internal/controller/dragonfly_controller.go
@@ -91,7 +91,7 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 		r.EventRecorder.Event(&df, corev1.EventTypeNormal, "Resources", "Created resources")
 		return ctrl.Result{}, nil
-	} else if df.Status.Phase == PhaseRollout {
+	} else if df.Status.IsRollingUpdate {
 		// This is a Rollout
 		log.Info("Rolling out new version")
 		var updatedStatefulset appsv1.StatefulSet
@@ -202,8 +202,8 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		// If we are here all are on latest version
 		r.EventRecorder.Event(&df, corev1.EventTypeNormal, "Rollout", "Completed")
 
-		// update status to ready
-		df.Status.Phase = PhaseReady
+		// update status
+		df.Status.IsRollingUpdate = false
 		if err := r.Status().Update(ctx, &df); err != nil {
 			log.Error(err, "could not update the Dragonfly object")
 			return ctrl.Result{Requeue: true}, err
@@ -242,7 +242,7 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 			// Start rollout and update status
 			// update status so that we can track progress
-			df.Status.Phase = PhaseRollout
+			df.Status.IsRollingUpdate = true
 			if err := r.Status().Update(ctx, &df); err != nil {
 				log.Error(err, "could not update the Dragonfly object")
 				return ctrl.Result{Requeue: true}, err
--- a/internal/controller/dragonfly_controller.go
+++ b/internal/controller/dragonfly_controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"time"
 
 	dfv1alpha1 "github.com/dragonflydb/dragonfly-operator/api/v1alpha1"
 	"github.com/dragonflydb/dragonfly-operator/internal/resources"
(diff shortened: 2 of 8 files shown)


internal/controller/dragonfly_controller.go
```diff
@@ -107,6 +116,81 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				// wait until the new replica is ready
+				// and completed a full sync
+				var newReplica corev1.Pod
+				if err := r.Get(ctx, client.ObjectKey{Namespace: df.Namespace, Name: replica.Name}, &newReplica); err != nil {
+					log.Error(err, "could not get new replica")
+					return ctrl.Result{}, err
+				}
+
+				// wait until the new replica has done a full sync
+				err := waitForFullSync(ctx, r.Client, &newReplica, 3*time.Minute)
```
ashotland: Do we have to wait here? 
Can we re-queue and check if we are done each time which i think is more inline with programming model and less error prone?
In any case I think that returning an error is problematic as we will not recover. 
So I think we should support  re-queuing the request and proceeding following a re-queue any how.

internal/controller/util.go
```diff
@@ -69,3 +72,62 @@ func isStatefulSetReady(ctx context.Context, c client.Client, name, namespace st
...
+				return err
+			}
+			if ready {
+				return nil
+			}
+		}
+	}
+}
+
+func isFullSync(ctx context.Context, c client.Client, pod *corev1.Pod) (bool, error) {
```
ashotland: It is not clear from the name if isFullSync means that full sync is in progress or done.

internal/controller/dragonfly_instance.go
```diff
@@ -198,10 +198,6 @@ func (dfi *DragonflyInstance) configureReplica(ctx context.Context, pod *corev1.
 		return err
 	}
 
-	if err := dfi.updateStatus(ctx, PhaseReady); err != nil {
```
Pothulapati: We should not update the status here as this might stop the rollout workflow in the middle. if its triggered.

internal/controller/dragonfly_controller.go
Pothulapati: Updated!

internal/controller/util.go
Pothulapati: I'm waiting on https://github.com/dragonflydb/dragonfly/issues/1132, Until then we can use this. I've updated the name also to match.

e2e/dragonfly_controller_test.go
```diff
@@ -208,7 +208,7 @@ var _ = Describe("Dragonfly Reconciler", Ordered, func() {
 		})
 
 		It("Update to image should be propagated successfully", func() {
-			newImage := resources.DragonflyImage + ":v0.17.0"
+			newImage := resources.DragonflyImage + ":v1.1.0"
```
ashotland, line 218: Update?

internal/controller/util.go
```diff
@@ -69,3 +92,69 @@ func isStatefulSetReady(ctx context.Context, c client.Client, name, namespace st
 
 	return false, nil
 }
+
+func waitForFullSync(ctx context.Context, c client.Client, pod *corev1.Pod, maxDuration time.Duration) error {
```
ashotland: remove?

internal/controller/dragonfly_controller.go
```diff
@@ -107,8 +228,31 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			// Start rollout and update status
+			// update status so that we can track progress
+			df.Status.Phase = PhaseRollout
+			if err := r.Status().Update(ctx, &df); err != nil {
+				log.Error(err, "could not update the Dragonfly object")
+				return ctrl.Result{Requeue: true}, err
+			}
+
+			// requeue so that the rollout is processed
+			return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```
ashotland, line 257: why wait 5 secs here?

internal/controller/dragonfly_controller.go
```diff
@@ -88,7 +89,127 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			log.Error(err, "could not list pods")
+			return ctrl.Result{Requeue: true}, err
+		}
+
+		if len(pods.Items) != int(*updatedStatefulset.Spec.Replicas) {
+			log.Info("Waiting for all replicas to be ready")
+			return ctrl.Result{Requeue: true}, nil
+		}
+
+		// filter replicas
```
ashotland: // filter replicas to master and replicas

internal/controller/dragonfly_controller.go
```diff
@@ -88,7 +89,127 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+					master = pod
+				}
+			} else {
+				log.Info("found pod without label", "pod", pod.Name)
+				// retry after they are ready
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+			}
+		}
+
+		// Check if all the latest replicas had a full sync
```
ashotland: // We want to update the replicas first then the master
// We want to have at most one updated replica in full sync phase at a time

internal/controller/dragonfly_controller.go
```diff
@@ -88,7 +89,127 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			// Check only with latest replicas
+			onLatestVersion, err := isPodOnLatestVersion(ctx, r.Client, &replica, &updatedStatefulset)
+			if err != nil {
+				log.Error(err, "could not check if pod is on latest version")
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, err
+			}
+			if onLatestVersion {
+				// check if the replica had a full sync
+				log.Info("New Replica found. Checking if replica had a full sync", "pod", replica.Name)
+				err := checkForStableSync(ctx, r.Client, &replica)
```
ashotland: I find checkForStableSync rerunning error only a bit confusing, it is not an error state to have a full sync in progress.
I'd go with something like
full_sync_in_progress = !checkForStableSync
...

if full_sync_in_progress {
  log.Info("Full sync in progress for an updated replica, requeuing...")
  return ctrl.Result{RequeueAfter: 5 * time.Second}
}

internal/controller/dragonfly_controller.go
```diff
@@ -88,7 +89,127 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				if err != nil {
+					log.Info("Not all new replicas are in stable sync yet", "pod", replica.Name, "reason", err)
+					return ctrl.Result{RequeueAfter: 5 * time.Second}, err
+				}
+				log.Info("Replica is in stable sync", "pod", replica.Name)
+				fullSyncedUpdatedReplicas++
+			}
+		}
+
+		log.Info(fmt.Sprintf("%d/%d are in stable sync", fullSyncedUpdatedReplicas, len(replicas)))
```
ashotland: stable state

internal/controller/util.go
```diff
@@ -69,3 +92,69 @@ func isStatefulSetReady(ctx context.Context, c client.Client, name, namespace st
...
+			err := checkForStableSync(ctx, c, pod)
+			if err != nil {
+				return err
+			}
+			return nil
+		}
+	}
+}
+
+func checkForStableSync(ctx context.Context, c client.Client, pod *corev1.Pod) error {
```
ashotland: checkForStableState

The replication phases are full-sync and stable-state

internal/controller/dragonfly_pod_lifecycle_controller.go
```diff
@@ -95,7 +95,7 @@ func (r *DfPodLifeCycleReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			}
 
 			r.EventRecorder.Event(dfi.df, corev1.EventTypeNormal, "Replication", "configured replication for first time")
-		} else if dfi.df.Status.Phase == PhaseReady {
+		} else if dfi.df.Status.Phase == PhaseReady || dfi.df.Status.Phase == PhaseRollout {
```
ashotland: Following our discussion - yea I think it will be better to introduce another status field IsRollingUpdateInProgress and then this controller won't have to deal with this aspect at all, right?

Pothulapati: Yep, Updated!

e2e/dragonfly_controller_test.go
Pothulapati: yes

internal/controller/dragonfly_controller.go
Pothulapati: We initially had a boolean state like you suggested, but with that as we check for multiple conditions inside `checkForStableState`, The condition here didn't really know what was wrong and it wasn't propogated in logs (as we don't pass a logger there). Hence, Changed it like this. It kind-off doesn't really matter if its an error state as we are handling it unlike a panic or something. Happy to update it back though if it makes sense.

internal/controller/dragonfly_controller.go
Pothulapati: Without a specific period, and just `Requeue: True`, It's taking a while to reconcile the object. 5 seconds is just a good enough timer. Should we go even lower? :thinking:

e2e/dragonfly_controller_test.go
ashotland: not sure if important but we are 1.2.1 by now :)

internal/controller/dragonfly_controller.go
```diff
@@ -88,7 +89,128 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			// Check only with latest replicas
+			onLatestVersion, err := isPodOnLatestVersion(ctx, r.Client, &replica, &updatedStatefulset)
+			if err != nil {
+				log.Error(err, "could not check if pod is on latest version")
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, err
+			}
+			if onLatestVersion {
+				// check if the replica had a full sync
+				log.Info("New Replica found. Checking if replica had a full sync", "pod", replica.Name)
+				err := checkForStableState(ctx, r.Client, &replica)
```
ashotland: I'd rather we change the checkForStableState signature from returning error or nil to return bool, err so we can distinguish between false and error and also I think is clearer to the reader.

internal/controller/dragonfly_controller.go
```diff
@@ -88,7 +89,128 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			if !onLatestVersion {
+				// delete the replica
+				log.Info("deleting replica", "pod", replica.Name)
+				r.EventRecorder.Event(&df, corev1.EventTypeNormal, "Rollout", "Deleting replica")
+				if err := r.Delete(ctx, &replica); err != nil {
+					log.Error(err, "could not delete pod")
+					return ctrl.Result{RequeueAfter: 5 * time.Second}, err
+				}
+
+				// Requeue and wait for it to be stable sync
```
ashotland: comment not relevant?

internal/controller/dragonfly_controller.go
```diff
@@ -88,7 +89,128 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			}
+			if onLatestVersion {
+				// check if the replica had a full sync
+				log.Info("New Replica found. Checking if replica had a full sync", "pod", replica.Name)
+				err := checkForStableState(ctx, r.Client, &replica)
+				if err != nil {
+					log.Info("Not all new replicas are in stable sync yet", "pod", replica.Name, "reason", err)
+					return ctrl.Result{RequeueAfter: 5 * time.Second}, err
+				}
+				log.Info("Replica is in stable sync", "pod", replica.Name)
```
ashotland: Shouldn't we delete here one old replica?
Currently, we need to provision the cluster for 2X resources for a rolling update, no?

Pothulapati: Why? New replica is only created when an old one is deleted. so why 2X? :thinking:

We delete the replica in the below code, Here we are just making sure that all the latest replicas are in stable state. If not we try again.

internal/controller/dragonfly_controller.go
Pothulapati: Updated

internal/controller/dragonfly_controller.go
ashotland: Sorry, fullSyncedUpdatedReplicas is just for logging purposes, I somehow imagined we wait for all of them :)

internal/controller/util.go
```diff
@@ -69,3 +90,51 @@ func isStatefulSetReady(ctx context.Context, c client.Client, name, namespace st
 
 	return false, nil
 }
+
+func checkForStableState(ctx context.Context, c client.Client, pod *corev1.Pod) (bool, error) {
```
ashotland: rename to isStableState now?

internal/controller/dragonfly_controller.go
ashotland: ok though I don't see any action take we need waiting for, in any case IIUC this happens once for every rolling update so not very important.

ashotland (approved):
LGTM - a couple of last tiny comments.

internal/controller/dragonfly_controller.go
Pothulapati: Yep!

PR #76, opened 2023-06-22
feat(resources): Add ability to set Annotations, Affinity and Tolerations

TheMerski:
We are running the dragonfly operator in our clusters, and missed some configuration options we would like for the Dragonfly pods.

I added `Annotations` which allows adding annotations.
We use [datadog Autodiscovery](https://docs.datadoghq.com/containers/kubernetes/integrations/?tab=kubernetesadv1#configuration) which works based on these annotations.

I also added `Tolerations` and `Affinity` options so we can configure on which nodes the pods run.

### DIFF
--- a/api/v1alpha1/dragonfly_types.go
+++ b/api/v1alpha1/dragonfly_types.go
@@ -41,11 +41,26 @@ type DragonflySpec struct {
 	// +kubebuilder:validation:Optional
 	Args []string `json:"args,omitempty"`
 
+	// (Optional) Annotations to add to the Dragonfly pods.
+	// +optional
+	// +kubebuilder:validation:Optional
+	Annotations map[string]string `json:"annotations,omitempty"`
+
 	// (Optional) Dragonfly container resource limits. Any container limits
 	// can be specified.
 	// +optional
 	// +kubebuilder:validation:Optional
 	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// (Optional) Dragonfly pod affinity
+	// +optional
+	// +kubebuilder:validation:Optional
+	Affinity *corev1.Affinity `json:"affinity,omitempty"`
+
+	// (Optional) Dragonfly pod tolerations
+	// +optional
+	// +kubebuilder:validation:Optional
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


manifests/dragonfly-operator.yaml
```diff
@@ -47,6 +47,839 @@ spec:
           spec:
             description: DragonflySpec defines the desired state of Dragonfly
             properties:
+              affinity:
```
Pothulapati: Can you remove the changes in this file? This is updated part of the release process. Updating it now, will cause problems to users who try to follow the README.

Pothulapati (commented):
Thanks for working on this! Will test and update my findings. Left a comment meanwhile.

manifests/dragonfly-operator.yaml
TheMerski: Yes, I have removed the changes from the commit.

TheMerski:
> Thanks for working on this! Will test and update my findings. Left a comment meanwhile.

Great, thanks for taking a look so fast!
It is my first time working on it so please let me know if there is something I need to change or update.

Pothulapati:
@TheMerski Sorry for the late get back but the change looks good! Can you also remove the changes in the `manifest/crd.yaml`? (Same reason as above. should have mentioned it earlier).

Once the CI passes, I will go ahead and merge it! :) Will probably do a Operator release this week to ship this change!

TheMerski:
> @TheMerski Sorry for the late get back but the change looks good! Can you also remove the changes in the `manifest/crd.yaml`? (Same reason as above. should have mentioned it earlier).
> 
> Once the CI passes, I will go ahead and merge it! :) Will probably do a Operator release this week to ship this change!

No problem, I have also removed the `crd.yaml` file now. 

It would be great if the new operator release can also contain at least dragonflydb v1.4.0 since we have been running these changes with v1.4.0 and noticed a massive latency decrease with v1.4.0.

Pothulapati:
> It would be great if the new operator release can also contain at least dragonflydb v1.4.0 since we have been running these changes with v1.4.0 and noticed a massive latency decrease with v1.4.0.


For sure! We will make sure to do that! :) Thanks again for the pull request.

Pothulapati (approved):
LGTM! Thanks again for the contribution! @TheMerski

Issue #50, opened 2023-04-17
resource: Add a way for the user to set redis password


Pothulapati:
Allow users to set a password for their dragonfly instance through a secret.
We should have to use the same secret from the Operator to run commands.

laurivosandi:
Make it possible to specify not only secret name but also the key inside the secret

Better yet have the operator populate the secret with various usable structures including the Redis URI

timtorChen:
How about supporting env and envFrom in Dragonfly CR? 
As setting up environment `DFLY_PASSWORD` is now possible https://github.com/dragonflydb/dragonfly/pull/792.

chakaz:
I can think of 2 ways to specify the password outside of the direct command line arguments
* Use `--flagfile` and set the password there
* Use `--fromenv` (or `--tryfromenv`), which allows specifying certain flags in env variables
For example:
```
FLAGS_requirepass=1234 dragonfly --fromenv=requirepass [...]
```

ksingh-scogo:
👀  much needed feature to use dragonfly that is deployed using operator in production setup

tekumara:
See https://www.dragonflydb.io/docs/managing-dragonfly/operator/authentication

chakaz:
For those interested in setting flags using environment variables, we recently added such support. See [here](https://www.dragonflydb.io/docs/managing-dragonfly/flags) for more details, but tl;dr is use `DFLY_flag` as env var and you should be good to go

Pothulapati:
Closing this as we support this through a secret now: https://www.dragonflydb.io/docs/managing-dragonfly/operator/authentication#password-based-authentication

Also, With `args` filed any extra number of configurations can also be passed.

PR #87, opened 2023-08-30
feat(resources): Updated the CRD to support setting environment variables and a service account

pstewy:
In order to get authentication working, the operator needed to be updated to communicate with the pods using their configured passwords. I also added support for setting an environment variable and service account, which allows for different methods of setting the password. I believe this satisfies some of the requests in https://github.com/dragonflydb/dragonfly-operator/issues/50

EDIT: since https://github.com/dragonflydb/dragonfly-operator/commit/4465e8082c05fafc34a3f029adba25a1ba8b4ad8, this PR does not modify how the operator communicates with pods.

### DIFF
--- a/api/v1alpha1/dragonfly_types.go
+++ b/api/v1alpha1/dragonfly_types.go
@@ -46,6 +46,11 @@ type DragonflySpec struct {
 	// +kubebuilder:validation:Optional
 	Annotations map[string]string `json:"annotations,omitempty"`
 
+	// (Optional) Env variables to add to the Dragonfly pods.
+	// +optional
+	// +kubebuilder:validation:Optional
+	Env []corev1.EnvVar `json:"env,omitempty"`
+
 	// (Optional) Dragonfly container resource limits. Any container limits
 	// can be specified.
 	// +optional
@@ -61,6 +66,11 @@ type DragonflySpec struct {
 	// +optional
 	// +kubebuilder:validation:Optional
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// (Optional) Dragonfly pod service account name
+	// +optional
+	// +kubebuilder:validation:Optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 }
 
 // DragonflyStatus defines the observed state of Dragonfly
--- a/internal/resources/resources.go
+++ b/internal/resources/resources.go
@@ -98,6 +98,7 @@ func GetDragonflyResources(ctx context.Context, df *resourcesv1.Dragonfly) ([]cl
 								},
 							},
 							Args: DefaultDragonflyArgs,
+							Env:  df.Spec.Env,
 							ReadinessProbe: &corev1.Probe{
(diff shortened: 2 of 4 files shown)


api/v1alpha1/dragonfly_types.go
```diff
@@ -46,6 +46,11 @@ type DragonflySpec struct {
 	// +kubebuilder:validation:Optional
 	Annotations map[string]string `json:"annotations,omitempty"`
 
+	// (Optional) Environment variables to add to the Dragonfly pods.
+	// +optional
+	// +kubebuilder:validation:Optional
+	Environment []corev1.EnvVar `json:"environment,omitempty"`
```
Abhra303: Let's use the traditional `Env` instead of `Environment` field.

internal/controller/util.go
```diff
@@ -138,3 +139,38 @@ func isStableState(ctx context.Context, c client.Client, pod *corev1.Pod) (bool,
...
+}
+
+// searchForPassword will search the pod spec for the redis password.
+// The password can be set in two ways:
+// 1. The env variable PasswordEnvironmentKey
+// 2. The argument PasswordArgumentKey
+// For each container, searchForPassword will look at the environment first, then the arguments,
+// and it will return on the first match it finds.
+func searchForPassword(pod *corev1.Pod) (string, bool) {
+	for _, container := range pod.Spec.Containers {
```
Abhra303: It doesn't make sense to search for the password in every container. Though we don't have sidecar containers for now, we should limit the scope to dragonfly containers only.

Add a `container.Name == "dragonfly"` check so that we only look into the dragonfly container env vars.

Abhra303 (commented):
Hi @pstewy, thanks for the PR! Left some comments :)

Abhra303:
The commit message doesn't have a header. Please use the PR title (without the trailing dot) as the commit message header. Also, the message description seems to have an indent on each line.

internal/controller/util.go
Abhra303: We don't need this function anymore as we are now using admin-port to prevent password or tls-enabled authentication to manage drangonfly instances by the operator.

pstewy: Awesome! That is a much better solution. I'll remove this.

internal/resources/resources.go
```diff
@@ -98,6 +98,7 @@ func GetDragonflyResources(ctx context.Context, df *resourcesv1.Dragonfly) ([]cl
 								},
 							},
 							Args: DefaultDragonflyArgs,
+							Env: df.Spec.Environment,
```
Abhra303: It should be `Env: df.Spec.Env` :)

Abhra303 (approved):

pstewy:
Hey @Abhra303 , thank you for accepting the pull request and for the prompt responses!
Sorry if I missed this in the documentation, but what is the release process for the operator? I'd love to use the new image with these changes.

Abhra303:
Hi @pstewy, a new Dragonfly version will be released soon (most probably this week). We will do a new release along with it.

PR #88, opened 2023-09-01
fix(resources): use cluster ip instead of headless svc

### LINKED ISSUE #84: Move to cluster IP service
Following https://discord.com/channels/981533931486724126/983406926782160896/threads/1144929396298362900

Headless service seems like a bad choice, it maps to the pod ip which keep changing and makes failover subject to dns cache delay.

We should switch so service IP which is static.
Our service matcher on the role label will make it hit the master pod.

@Pothulapati, @Abhra303 WDYT?

Abhra303:
Fixes #84


Pothulapati (approved):
LGTM

PR #90, opened 2023-09-04
fix: use admin-port for dragonfly operations

Abhra303:
Dragonfly operator should use admin port to manage operations on TLS or password enabled dragonfly instances.

### DIFF
--- a/e2e/dragonfly_controller_test.go
+++ b/e2e/dragonfly_controller_test.go
@@ -94,7 +94,7 @@ var _ = Describe("Dragonfly Reconciler", Ordered, func() {
 			// check resource requirements of statefulset
 			Expect(ss.Spec.Template.Spec.Containers[0].Resources).To(Equal(*df.Spec.Resources))
 			// check args of statefulset
-			Expect(ss.Spec.Template.Spec.Containers[0].Args[1:]).To(Equal(df.Spec.Args))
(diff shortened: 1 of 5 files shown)


internal/resources/const.go
```diff
@@ -20,6 +20,9 @@ const (
 	// DragonflyPort is the port on which Dragonfly listens
 	DragonflyPort = 6379
 
+	// DragonflyAdminPort is the admin port on which Dragonfly listens
```
ashotland, line 25: // IMPORTANT: This port should not be opened to non trusted networks.

e2e/dragonfly_controller_test.go
```diff
@@ -94,7 +94,7 @@ var _ = Describe("Dragonfly Reconciler", Ordered, func() {
 			// check resource requirements of statefulset
 			Expect(ss.Spec.Template.Spec.Containers[0].Resources).To(Equal(*df.Spec.Resources))
 			// check args of statefulset
-			Expect(ss.Spec.Template.Spec.Containers[0].Args[1:]).To(Equal(df.Spec.Args))
+			Expect(ss.Spec.Template.Spec.Containers[0].Args[2:]).To(Equal(df.Spec.Args))
```
ashotland: Can we make the default args some const and then concat here so we don't need to skip them and maintain the test in sync with the number of args? 
(or any other way?)

ashotland (approved):

PR #91, opened 2023-09-04
feat(tls): Support for Server TLS

### LINKED ISSUE #103: feature: Server Side TLS in Dragonfly

Abhra303:
Fixes #103 

Supports Server Side TLS for Dragonfly. This PR adds a way to read TLS secrets for the Dragonfly connections to use. With this the Clients, have to make sure to have the relevant `ca.crt` to verify the Dragonfly connection. As we don't support (client TLS yet as we don't pass --tls-ca-crt yet), Passing a password is mandated right now. Once we have client TLS verification, Then it would be either password or client TLS. Relevant checks will be added.

After this, If a Secret is set Dragonfly will use those certs for secure communication and expects at least one auth mechanism from the client. On the client side, It needs to trust the Certificate (and hence the usage of ca.crt in docs) to establish the communication.


Pothulapati:
Heads up, Taking this PR from Abhra as this is a bit more complicated and requires cert-manager integration with docs.

ashotland (approved):

Issue #86, opened 2023-08-28
Support for ARM architecture


monlor:
Please consider adding support for the ARM architecture to enable usage on ARM-based server. Thank you!

DarkmatterVale:
+1 this comment, would like to run Dragonfly on a Kubernetes cluster with ARM nodes.

Not sure if this is too much work, given that it's already possible to run DragonflyDB itself on an ARM host.

Pothulapati:
@DarkmatterVale You are right, and it isn't too much work. Considering that we already have a PR for this, You can expect this to be available in the coming days!

Pothulapati:
Closing as this is fixed #126

PR #92, opened 2023-09-05
feat: 🎸 Support arm image building

monlor:
Fixes https://github.com/dragonflydb/dragonfly-operator/issues/86

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: docker.dragonflydb.io/dragonflydb/operator
-  newTag: v0.0.6
+  newTag: v0.0.7
--- a/manifests/crd.yaml
+++ b/manifests/crd.yaml
@@ -873,6 +873,115 @@ spec:
                 items:
                   type: string
                 type: array
+              env:
+                description: (Optional) Env variables to add to the Dragonfly pods.
+                items:
+                  description: EnvVar represents an environment variable present in
+                    a Container.
+                  properties:
+                    name:
+                      description: Name of the environment variable. Must be a C_IDENTIFIER.
+                      type: string
+                    value:
+                      description: 'Variable references $(VAR_NAME) are expanded using
+                        the previously defined environment variables in the container
+                        and any service environment variables. If a variable cannot
+                        be resolved, the reference in the input string will be unchanged.
+                        Double $$ are reduced to a single $, which allows for escaping
+                        the $(VAR_NAME) syntax: i.e. "$$(VAR_NAME)" will produce the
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -49,7 +49,7 @@ jobs:
     needs: build-operator
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
     env:
-      IMG: ghcr.io/dragonflydb/operator:${{ github.ref_name }}
+      IMG: ghcr.io/${{ github.actor }}/dragonfly-operator:${{ github.ref_name }}
```
Abhra303: I don't think we need to change `IMG` or `tags` for `release-operator` job.

monlor: Maybe I think this is more conducive to the development of the fork project. If you use the operator image name, such as monlor/operator, it will be easy to confuse.

Abhra303: But it is not needed for the `release-operator` job. This job runs only when a new version is released, not when a pull request is created.

monlor: Okay, I restored the image name.

.github/workflows/ci.yml
```diff
@@ -68,12 +74,13 @@ jobs:
...
 
       - name: Build and Publish image into GHCR
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v4
         with:
           context: .
           file: ./Dockerfile
           push: true
-          tags: ghcr.io/dragonflydb/operator:${{ github.ref_name }}
+          tags: ghcr.io/${{ github.actor }}/operator:${{ github.ref_name }}
```
Abhra303: you need to revert this line. Because our operator images are released under the `ghcr.io/dragonflydb/operator` name and it doesn't make sense to use `${{ github.actor }}` here.

.github/workflows/ci.yml
```diff
@@ -90,5 +97,5 @@ jobs:
           body: |
             Release ${{ github.ref_name }}
 
-            Docker image: ghcr.io/dragonflydb/operator:${{ github.ref_name }}
+            Docker image: ghcr.io/${{ github.actor }}/operator:${{ github.ref_name }}
```
Abhra303: same.

.github/workflows/ci.yml
Abhra303: You need to revert this line (and consecutive lines containing `ghcr.io/${{ github.actor }}/operator`) in order to make it work.

Pothulapati:
Was trying to push the changes to your branch but somehow it got closed (and can't reopen)? So sorry @monlor. I'll wait a day before I raise a PR but essentially your changes are good and if you can set the actor as `dragonflydb` in the release job, I'm happy to merge.

PR #94, opened 2023-09-11
fix(e2e): add tests for latest features

Pothulapati:


internal/resources/resources.go
```diff
@@ -96,6 +96,10 @@ func GetDragonflyResources(ctx context.Context, df *resourcesv1.Dragonfly) ([]cl
 									Name:          DragonflyPortName,
 									ContainerPort: DragonflyPort,
 								},
+								{
+									Name:          "admin",
```
Abhra303, line 100: Nice catch 👍

Abhra303 (approved):
LGTM

internal/resources/resources.go
Pothulapati: This wasn't the issue actually i.e https://stackoverflow.com/questions/63526825/kubernetes-service-without-defined-port#:~:text=Yes%20you%20can%20expose%20service,uses%2C%20but%20is%20primarily%20informational.

But the issue was to have `admin_nopass` otherwise when password is set, we get a bunch of noauth errors. This is required.

PR #101, opened 2023-09-27
feat(monitoring): add guide to use PodMonitor with operator

Abhra303:


internal/resources/const.go
```diff
@@ -67,7 +67,6 @@ const (
 
 var DefaultDragonflyArgs = []string{
 	"--alsologtostderr",
-	"--primary_port_http_enabled=false",
```
Pothulapati: why? better to scrape from admin so that we separate all these tasks?

Abhra303: In that case, we need an admin service, right?

Pothulapati: Why? 

Looking at the service monitor, we are scraping the pods using the selectors right? changing the port to be `9999` should do it. Can you check if that works?

Abhra303: serviceMonitors uses services, so as far as I understand, it targets services and scrap data through it. I am checking though

monitoring/podMonitorGuide.md
```diff
@@ -0,0 +1,96 @@
...
+
+```
+LATEST=$(curl -s https://api.github.com/repos/prometheus-operator/prometheus-operator/releases/latest | jq -cr .tag_name)
+curl -sL https://github.com/prometheus-operator/prometheus-operator/releases/download/${LATEST}/bundle.yaml | kubectl create -f -
+```
+
+## Step 2:
+Now that we have installed the operator, we can create `prometheus` resources. If you have RBAC enabled, create necessary `serviceaccount`, `clusterrole` and `clusterrolebinding` resources first.
+
+ServiceAccount -
```
Pothulapati: Can you switch to commands here directly? to make it more usable and tryable? like we have in the readme?

Abhra303: Ok sure

monitoring/podMonitor.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: dragonfly-monitor
```
Pothulapati: because these pod monitors are specific to the `dragonfly-sample` DF instance, I'd advice on having the same name i.e `dragonfly-sample` instead of generic naming

Pothulapati (approved):
LGTM and works as expected.

PR #109, opened 2023-09-28
feat(upgrade): Use repltakeover

Pothulapati:
Instead of deleting the master, This PR updates the rollout flow to run `repltakeover` on a latest replica which would mark this pod as the master while terminating the old master. When a pod is terminated, Kubernetes automatically recreates it and the pod lifecycle controller will do the needful.

The pod lifecycle controller does react on various instances and tries to corrupt, but we only expect it to handle new pod creations and hence toggle is added to not react on `isRollingUpdate=true` for other events.

### DIFF
--- a/e2e/util.go
+++ b/e2e/util.go
@@ -99,7 +99,7 @@ func checkAndK8sPortForwardRedis(ctx context.Context, clientset *kubernetes.Clie
 		return nil, fmt.Errorf("no master pod found")
 	}
 
-	fw, err := portForward(ctx, clientset, config, master, stopChan, resources.DragonflyAdminPort)
+	fw, err := portForward(ctx, clientset, config, master, stopChan, resources.DragonflyPort)
 	if err != nil {
 		return nil, err
 	}
--- a/e2e/util.go
+++ b/e2e/util.go
@@ -105,7 +105,7 @@ func checkAndK8sPortForwardRedis(ctx context.Context, clientset *kubernetes.Clie
 	}
 
 	redisOptions := &redis.Options{
-		Addr: fmt.Sprintf("localhost:9998"),
+		Addr: fmt.Sprintf("localhost:%d", resources.DragonflyPort),
 	}
 
 	if password != "" {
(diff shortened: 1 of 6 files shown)


internal/controller/util.go
```diff
@@ -56,6 +57,57 @@ func isPodOnLatestVersion(ctx context.Context, c client.Client, pod *corev1.Pod,
...
+		LabelSelector: labels.SelectorFromValidatedSet(map[string]string{
+			"app":                              statefulSet.Name,
+			resources.KubernetesPartOfLabelKey: "dragonfly",
+		}),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Iterate over the pods and find the one which is on the latest version
```
ashotland: finds a one (could be more than one right?)

ashotland (approved):
LGTM with minor comment

internal/controller/util.go
Pothulapati: Yep, and usually the case as we start from replicas and come here. here, we choose the first latest replica which would be the least numbered

ashotland: so I'd modify the comment a bit "find the one" -> find a one

ashotland (approved):

e2e/util.go
```diff
@@ -99,7 +99,7 @@ func checkAndK8sPortForwardRedis(ctx context.Context, clientset *kubernetes.Clie
 		return nil, fmt.Errorf("no master pod found")
 	}
 
-	fw, err := portForward(ctx, clientset, config, master, stopChan, resources.DragonflyAdminPort)
+	fw, err := portForward(ctx, clientset, config, master, stopChan, resources.DragonflyPort)
```
Abhra303, line 102: I see `portForward` doesn't use the last parameter (port). So there is no impact of this line change. Maybe update the portForward definition?

Pothulapati: That was an unnecessary change missed in the old PR I think. Changed it back to use the port parameter

PR #111, opened 2023-10-02
feat(e2e): add utilities to run commands on cluster from test

Abhra303:


e2e/util.go
```diff
@@ -61,3 +77,98 @@ func isStatefulSetReady(ctx context.Context, c client.Client, name, namespace st
...
+		return nil, err
+	}
+
+	// create the clientset
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	pods, err := clientset.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{
```
Pothulapati: pass ctx?

Pothulapati (approved):
LGTM apart from that comment.

e2e/util.go
Abhra303: Good catch 👍

PR #118, opened 2023-10-10
fix(e2e): cleanup after each test

Pothulapati:


e2e/dragonfly_controller_test.go
```diff
@@ -472,6 +472,18 @@ var _ = Describe("Dragonfly Lifecycle tests", Ordered, func() {
...
+
+		It("Cleanup", func() {
+			var df dragonflydbiov1alpha1.Dragonfly
+			err := k8sClient.Get(ctx, types.NamespacedName{
+				Name:      name,
+				Namespace: namespace,
+			}, &df)
+			Expect(err).To(BeNil())
+
+			err = k8sClient.Delete(ctx, &df)
```
Abhra303, line 483: Should we also check if the related resources are also deleted (i.e. deployments, service etc)?

e2e/dragonfly_pod_lifecycle_controller_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	dfv1alpha1 "github.com/dragonflydb/dragonfly-operator/api/v1alpha1"
+	dragonflydbiov1alpha1 "github.com/dragonflydb/dragonfly-operator/api/v1alpha1"
```
Abhra303: Why do we need the same import with different import name (`dfv1alpha1` and `dragonflydbiov1alpha1`?

Abhra303 (approved):
LGTM with some minor comments

e2e/dragonfly_controller_test.go
Pothulapati: Deletion is not the purview of the Operator. It is done automatically by the garbage collector in k8s. This is through the Owner reference that we pass to resources.

Abhra303: Agreed

PR #134, opened 2023-11-17
feat(resources): Add labels field to spec, which will be added to pods

WesleyKlop:
Hey there 👋 

We wanted to add labels to the dragonfly pods which are used by some other tooling we have. Since that was not possible before, this PR adds support for that!

Please let me know if I'm missing something. One of the considerations I had is that the operator already assigns some labels to the pods/statefulset so we should not allow the user to override them (for now?).


Abhra303 (approved):
Hi @WesleyKlop, thanks for the PR. LGTM!

Issue #133, opened 2023-11-17
Setting admin_nopass to false prevents operator from configuring instances


diffuse:
Configuring authentication on the non-admin port works perfectly with:
```yaml
authentication:
  passwordFromSecret:
    name: foo
    key: password
```

However, on inspection of `kubectl describe statefulset dragonfly`, containers are configured with the following arguments:
```text
Args:
      --alsologtostderr
      --primary_port_http_enabled=false
      --admin_port=9999
      --admin_nopass
```

I confirmed that when authentication is configured, connecting to the non-admin port requires a password, and connecting to the admin port does not.  I assumed that overriding `admin_nopass` would cause the operator to authenticate with the secret above, but it still attempts an unauthenticated connection to the admin port.

```text
2023-11-17T02:39:39Z    ERROR   Failed to mark pod as master    {"controller": "pod", "controllerGroup": "", "controllerKind": "Pod", "Pod": {"name":"dragonfly-0","namespace":"default"}, "namespace": "default", "name": "dragonfly-0", "reconcileID": "<id>", "podName": "dragonfly-0", "error": "error running SLAVE OF NO ONE command: NOAUTH Authentication required."}
github.com/dragonflydb/dragonfly-operator/internal/controller.(*DragonflyInstance).configureReplication
        /workspace/internal/controller/dragonfly_instance.go:108
github.com/dragonflydb/dragonfly-operator/internal/controller.(*DfPodLifeCycleReconciler).Reconcile
        /workspace/internal/controller/dragonfly_pod_lifecycle_controller.go:92
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.4/pkg/internal/controller/controller.go:122
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.4/pkg/internal/controller/controller.go:323
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.4/pkg/internal/controller/controller.go:274
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
[...]

PR #135, opened 2023-11-18
feat(auth): Support operator management of instances with password protected admin ports

diffuse:
This PR allows the operator to manage instances of dragonfly that have the argument `--admin_nopass=false` and authentication enabled. 

There are updates to the README included, but in short: a secret in the namespace `dragonfly-operator-system` must be created in the form:
```yaml
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: dragonfly-admin-secret
  namespace: dragonfly-operator-system
data:
  password: c29tZS1wYXNzd29yZA==
```

Which the operator will use to open authenticated admin connections and to set `masterauth` on replicas.

I'm having a bit of trouble running all of the tests on minikube (some appear to hang, etc).  I'd like to add some new tests for this feature, but since I can't run all of the originals successfully, I'm opening this PR and asking for advisement.  

Let me know what you think/if you see any issues.  Thanks for this project!

(this addresses https://github.com/dragonflydb/dragonfly-operator/issues/133)


diffuse:
From the conflicts, l'm not quite sure what happened.  I edited only a few files.  Please let me know what I need to change to fix the PR.  Thanks!

PR #136, opened 2023-11-18
feat(auth): Support operator management of instances with password protected admin ports

diffuse:
This PR allows the operator to manage instances of dragonfly that have the argument `--admin_nopass=false` and authentication enabled. 

There are updates to the README included, but in short: a secret in the namespace `dragonfly-operator-system` must be created in the form:
```yaml
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: dragonfly-admin-secret
  namespace: dragonfly-operator-system
data:
  password: c29tZS1wYXNzd29yZA==
```

Which the operator will use to open authenticated admin connections and to set `masterauth` on replicas.

I'm having a bit of trouble running all of the tests on minikube (some appear to hang, etc).  I'd like to add some new tests for this feature, but since I can't run all of the originals successfully, I'm opening this PR and asking for advisement.  

Let me know what you think/if you see any issues.  Thanks for this project!

(this addresses https://github.com/dragonflydb/dragonfly-operator/issues/133)


diffuse:
Hello, I don't mean to be a bother, but is there any news/discussion on this PR?  Thanks! :smile:

README.md
```diff
@@ -83,7 +83,47 @@ kubectl patch dragonfly dragonfly-sample --type merge -p '{"spec":{"resources":{
...
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: dragonfly-admin-secret
+  namespace: dragonfly-operator-system
+data:
+  password: c29tZS1wYXNzd29yZA==
+```
+The dragonfly operator will use this secret to authenticate on the admin port of dragonfly pods so it can run admin commands.  It will also use this password to allow replicas to connect to the master on the authenticated admin port.  A duplicate secret must be created in the `dragonfly-operator-system` namespace because the operator will usually run in a namespace separate from the dragonfly pods, and kubernetes disallows access to cross-namespace secrets.
```
Pothulapati: Hmm, So We create a admin secret and pass it to the Operator, but how are even the Dragonfly instances admin secret configured? manually by the user? 🤔

Pothulapati (commented):
Sorry for the late reply but tyring to understand more.

internal/controller/dragonfly_instance.go
```diff
@@ -340,9 +362,19 @@ func (dfi *DragonflyInstance) getPods(ctx context.Context) (*corev1.PodList, err
...
-		Addr: fmt.Sprintf("%s:%d", pod.Status.PodIP, resources.DragonflyAdminPort),
+		Addr:     fmt.Sprintf("%s:%d", pod.Status.PodIP, resources.DragonflyAdminPort),
+		Username: dfi.dfAdminCreds.username,
+		Password: dfi.dfAdminCreds.password,
 	})
 
+	// if admin credentials are defined, set masterauth so slave
+	// can talk to master over an authenticated connection
+	if dfi.dfAdminCreds.password != "" {
+		if err := redisClient.ConfigSet(ctx, "masterauth", dfi.dfAdminCreds.password).Err(); err != nil {
```
Pothulapati: Ahh, So this is where we set to instances but I don't think setting here makes sense. We should try to pass it during initialisation through args directly.

Pothulapati: Hmm. I don't think we are even setting this here. we are asking the password to be used but where are we even configuring the master with the admin password?

I think, the current approach looks a bit complicated. I'd rather prefer somehow letting the user give access to the password secret to the Operator and it using it. instead of this. wdyt?

diffuse: Hi @Pothulapati , thanks for getting back to me! :)

Yes, this requires that the admin password provided to the operator via k8s secret match the password provided to the actual dragonfly instances.  I absolutely agree that sharing the secret between the instances and operator would be better, but I don't see a way around this, since secrets can't be shared across k8s namespaces.  

We'd either have to enforce that the operator and dragonfly instances exist in the same namespace, which would require a major change to `dragonfly-operator.yaml`, or require that the users create a duplicate secret in the `dragonfly-operator-system` that matches the instance secret (like I did in this PR).  Do you know of another way to share the k8s secret?

I figured setting the `masterauth` in replica configuration made sense, since it's an implementation detail tied to the configuration of a replica, but I'm open to moving it elsewhere.

README.md
diffuse: When using `passwordFromSecret`, the dragonfly instances will have both the `6379` port password and the admin password set to the password in the below secret:
```yaml
spec:
  authentication:
    passwordFromSecret:
      name: dragonfly-secret
      key: password
```

The only extra thing necessary to enable the use of this password on the admin port is setting `admin_nopass`:
```yaml
spec:
  args:
    - --admin_nopass=false
```

So, as long as this secret matches the secret in the `dragonfly-operator-system` namespace, authenticated admin connections can be used.

internal/controller/dragonfly_instance.go
Pothulapati: > Do you know of another way to share the k8s secret?

Considering that we are on a operator, Nothing should prevent us from querying secrets in other namespaces (just like we are creating resources in other namespaces) given our rbac was given those permissions. wdyt?

diffuse: Ok, if we have the appropriate permissions to read secrets my approach would probably be to just read the secret defined in `passwordFromSecret` when creating dragonfly instances and use that as the admin/masterauth password.  I remember trying this initially and running into a permissions issue when reading the secret, but I'll give it another shot and see if I can get it working.  How does that sound?

Pothulapati: Sounds good! Did you have any blocker there @diffuse ? You can create the relevant rbac to get the operator access to the secret, and that should be it.

diffuse: Thanks!  No blocker, I just haven't had any time to get to this yet.  I should be able to implement it soon though.

diffuse: I finally found some time today and implemented most of this feature.  Once I run a few more tests I'll push it for review hopefully within the next 3 days.  Thanks for the patience!

diffuse: @Pothulapati I implemented the changes.  Let me know what you think.  Thanks!

Pothulapati: Thanks @diffuse! Taking a look today!

diffuse: Please let me know if any further explanation would help.

Pothulapati:
@diffuse First of all really sorry for the all the waiting. This is especially hard for issues like this as its hard to have a balance!

While we agree reading the secret is the answer to this, We aren't sure if this would be the right for everyone! I'm sure some users would not like giving the Operator access to all the secrets in the cluster and would not deploy the Operator in this case. We can add a toggle but we aren't yet sure how many users actually want this as most users trust applications in their cluster and many others who have issues like this probably already have network RBAC around which services can access what. Until we have more requests from others users, We are not sure if its the right thing to go ahead on this. :/ Closing this hence.

PR #146, opened 2024-01-30
Update resources.go

zaphinath:
Change the directory to match the location of the ca.crt for the healthcheck when tls is enabled.


Pothulapati (approved):
LGTM

PR #148, opened 2024-01-31
feat(crd): add AclFromSecret field

### LINKED ISSUE #144: add support to pass in a aclfile
We have the ability to use the `--aclfile` option as an `arg` but no way to inject this file.   A [secret mounted through a volume](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#create-a-pod-that-has-access-to-the-secret-data-through-a-volume) would work.  I'm also open to other options.  

Thanks for this awesome project!

Abhra303:
Closes #144


Pothulapati (approved):
LGTM

Issue #168, opened 2024-03-19
Helm chart


onedr0p:
Hi 👋🏼 

It would be awesome if Helm was a supported method to install the operator and CRDs.

Thanks!

nujragan93:
would like to help with this /assign

TheDukeDK:
Would it be possible to ALSO package the chart in the old format along with the OCI formatted chart and push it to https://artifacthub.io?

It would be helpful for expanding adoption of the operator I think. We, for example, have a whole crap load of scripting around helm where we use the old way of doing `helm repo add`. 

I guess there is at least someone else that also has this need as they seem to have done this. See [Jacob Colvin's Helm Charts](https://artifacthub.io/packages/helm/jacobcolvin/dragonfly-operator).

PR #174, opened 2024-04-06
feat/helmChart: initial support for installing the operator using helm charts

nujragan93:
Hey there 👋

This PR adds initial support for helm chart
 https://github.com/dragonflydb/dragonfly-operator/issues/168


charts/dragonfly-operator/templates/namespace.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.createOperatorNamespace -}}
+---
+apiVersion: v1
+kind: Namespace
```
onedr0p: I don't think we should be including a namespace resource, helm lets you install into namespaces via 

```helm install --create-namespace --namespace dragonfly ...```

and that will set all the resources to that namespace in the templates.

Safe to also say I do not see including the namespace resource being done in other helm charts as well.

**Comment by onedr0p** [comment]
I don't see any workflow changes to publish the Helm chart (e.g. https://github.com/dragonflydb/dragonfly/blob/b994f934d3311925d4dc48dcb87f5c560d478b27/.github/workflows/docker-release.yml#L82-L96)

File: charts/dragonfly-operator/values.yaml [hunk 2]
```diff
@@ -0,0 +1,128 @@
...
+    tag: v1.1.2
+
+
+  resources:
+    limits:
+      cpu: 500m
+      memory: 128Mi
+    requests:
+      cpu: 10m
+      memory: 64Mi
```
onedr0p:
```suggestion
  resources: {}
```

Resources should not have a default

charts/dragonfly-operator/values.yaml
```diff
@@ -0,0 +1,128 @@
...
+                operator: In
+                values:
+                  - amd64
+                  - arm64
+                  - ppc64le
+                  - s390x
+              - key: kubernetes.io/os
+                operator: In
+                values:
+                  - linux
```
onedr0p:
```suggestion
  affinity: {}
```

I don't see why there needs to be a default set for the affinity. Better to leave this unset and leave it up to the user.

charts/dragonfly-operator/values.yaml
nujragan93: missed commenting out those! thanks

charts/dragonfly-operator/values.yaml
nujragan93: missed commenting out those! thanks

nujragan93:
> Thanks for tackling this @nujragan93 I don't see any workflow changes to publish the Helm chart (e.g. what is done in https://github.com/dragonflydb/dragonfly/blob/main/.github/workflows/docker-release.yml)

that seems like a OCI repo for dragonflyDb, does the operator need to have a separate repo ? do we need to add tests to the CI before pushing ?

onedr0p:
> that seems like a OCI repo for dragonflyDb, does the operator need to have a separate repo ?

They seem to be packaging up the Helm chart into a [OCI artifact](https://helm.sh/blog/storing-charts-in-oci/) which seems like that could be done here as well and published under https://github.com/orgs/dragonflydb/packages?repo_name=dragonfly-operator

> do we need to add tests to the CI before pushing ?

It looks like they have some CI tests in https://github.com/dragonflydb/dragonfly/tree/main/contrib/charts/dragonfly/ci so probably might be a good idea to have some here as well.

charts/dragonfly-operator/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+    capabilities:
+      drop:
+        - ALL
+
+manager:
+  image:
+    repository: docker.dragonflydb.io/dragonflydb/operator
+    pullPolicy: IfNotPresent
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: v1.1.2
```
onedr0p:
```suggestion
    tag: ""
```

This should come from the Chart.yaml `appVersion` field as implemented below

charts/dragonfly-operator/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
onedr0p:
```suggestion
appVersion: "v1.1.2"
```

This is the current operator version

charts/dragonfly-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,96 @@
...
+        - name: manager
+          args:
+            - --health-probe-bind-address=:8081
+            - --metrics-bind-address=127.0.0.1:8080
+            - --leader-elect
+          command:
+            - /manager
+          securityContext:
+            {{- toYaml .Values.manager.securityContext | nindent 12 }}
+          image: "{{ .Values.manager.image.repository }}:{{ .Values.manager.image.tag  }}"
```
onedr0p:
```suggestion
          image: "{{ .Values.manager.image.repository }}:{{ .Values.manager.image.tag | default .Chart.AppVersion  }}"
```

Should default to the Chart.yaml `appVersion`

charts/dragonfly-operator/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+  securityContext:
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop:
+        - ALL
+
+manager:
+  image:
+    repository: docker.dragonflydb.io/dragonflydb/operator
```
onedr0p, line 65:
```suggestion
    repository: ghcr.io/dragonflydb/operator
```

onedr0p: Not sure if GHCR is better since it is published here?

charts/dragonfly-operator/templates/crds.yaml
```diff
@@ -0,0 +1,1611 @@
+{{- if .Values.crds.install }}
```
onedr0p, line 1: There's a long lengthy debate on how CRDs should be managed in Helm charts. Either by what you did here, or dropping them un-templated in a `crds` folder and letting helm install them by default. If the dragonfly team is okay with this method then it shouldn't be a problem moving forward.

I honestly prefer the `crds` folder method they describe [here](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) but there are some caveats. The main one being the CRDs will not be upgraded even if you change them here in the chart. There are ways around that... Flux/Argo natively supports upgrading the chart as well as the CRDs while other CI/CD tools might have to `kubectl apply` the CRDs while upgrading the Chart.

charts/dragonfly-operator/values.yaml
nujragan93: @Abhra303 @Pothulapati any ideas?

Pothulapati:
Hey @nujragan93, Thanks for working on this.

+1 to both of these questions. We can have similar CI linting, generation tests like Dragonfly. Feel free to add a simpler verison of that. 

Yep, We should be using Github Packages to store and publish the Helm Chart.

(Reviewing the Helm Chart itself meanwhile)

charts/dragonfly-operator/values.yaml
Pothulapati: We should use docker.dragonflydb.io. It is just a gateway to ghcr, so its all the same.

nujragan93:
> Hey @nujragan93, Thanks for working on this.
> 
> +1 to both of these questions. We can have similar CI linting, generation tests like Dragonfly. Feel free to add a simpler verison of that.
> 
> Yep, We should be using Github Packages to store and publish the Helm Chart.
> 
> (Reviewing the Helm Chart itself meanwhile)

I have made some workflow changes to push helm chart to github(dont know how to test it)
This PR seems already very big, I feel the ci linting and tests should be made a separate PR.

onedr0p:
@nujragan93 pipelines looks good. I also suggest we separate the PRs for adding ci linting and tests after this is merged.

nujragan93:
> @nujragan93 pipeline looks good. I agree we should separate the PRs for adding ci linting and tests after this is merged.

agreed.

charts/dragonfly-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,96 @@
...
+      serviceAccountName: {{ include "dragonfly-operator.serviceAccountName" . }}
+      terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - args:
+            - --secure-listen-address=0.0.0.0:8443
+            - --upstream=http://127.0.0.1:8080/
+            - --logtostderr=true
+            - --v=0
```
onedr0p, line 38: It would be nice to allow for extra args here.

charts/dragonfly-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,96 @@
...
+              memory: 128Mi
+            requests:
+              cpu: 5m
+              memory: 64Mi
+
+        - name: manager
+          args:
+            - --health-probe-bind-address=:8081
+            - --metrics-bind-address=127.0.0.1:8080
+            - --leader-elect
```
onedr0p, line 60: It would be nice to allow for extra args here.

onedr0p:
@Pothulapati any update here?

onedr0p:
@Pothulapati, or anybody?

.github/workflows/ci.yml
```diff
@@ -73,6 +73,30 @@ jobs:
...
+
+      - name: Update helm chart
+        run: |
+          git checkout main
+          sed -Ei \
+           -e 's/^(version\:) .*/\1 '${{ env.VERSION }}'/g' \
+           -e 's/^(appVersion\:) .*/\1 "'${{ env.VERSION }}'"/g' \
+           charts/dragonfly-operator/Chart.yaml          
+          
+          git commit \
```
Pothulapati, line 87: nit: Is there a need to commit the version into repository? considering we publish the chart as an artifact? :thinking:

charts/dragonfly-operator/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
Pothulapati, line 1: Not really sure what this file is for :thinking:

Pothulapati (commented):
Added some nits and questions! Code looks mostly good!

Considering Helm is very errror prone (easy to add formatting errors), We should be adding linting tests like https://github.com/dragonflydb/dragonfly/blob/main/.github/actions/lint-test-chart/action.yml, Otherwise we have no way to make sure these charts actually work. This is very important considering we have so many values, etc.

We can merge it once we see that these tests pass.

Thanks! Really sorry for missing the review on this.

charts/dragonfly-operator/templates/tests/test-connection.yaml
onedr0p: https://helm.sh/docs/topics/chart_tests/

.github/workflows/ci.yml
onedr0p: It's also being done [here](https://github.com/dragonflydb/dragonfly/blob/6b8a8b6410c84f0d03444a6700f54fb806270253/.github/workflows/docker-release.yml#L93) in dragonflydb/dragonfly, so points for consistency? 😆

onedr0p:
Anything I can do here to help move this along @Pothulapati @nujragan93 ?

.github/workflows/ci.yml
Pothulapati: Sure!

Pothulapati:
@onedr0p Feel free to  create a new branch from this branch and add your commits with the changes that I requested i.e test the chart. I'm happy to follow up with a review.

Pothulapati:
>  @Pothulapati you are a maintainer of this repo which gives you the ability to edit this PR with those said changes. 

Yes, but that does not mean I have enough bandwidth to work on everything :/

> If that's not something you want to do I'll go ahead and create a branch off this and open a PR with your requested changes.

If you are interested, and want to help, Feel free!

onedr0p:
@Pothulapati  After looking this over... there's quite a bit of CI changes that need to be done to support this. Why is it not possible to merge this PR and iterate on the changes in future updates?

nujragan93:
> @Pothulapati After looking this over... there's quite a bit of CI/GitHub workflow changes that need to be done to support what you ask because the workflow you have in this repo is only for releases / pushing to main. There is no testing workflow to expand upon like here https://github.com/dragonflydb/dragonfly/tree/main/.github/workflows
> 
> Why is it not possible to merge this PR and iterate on the changes in future updates? As @nujragan93 already mentioned this PR is pretty big already.

I agree, if the workflows are setup, then I would jump in and iterate on tests with the helm charts.

Abhra303 (approved):

Pothulapati:
Sorry folks but this is merged now! Looking forward to the next iterations!

FlorisFeddema:
Does this need a new operator release first? Seems like the package is not yet available in the ghcr

Pothulapati:
Just FYI, we had to remove the release changes as it wasn't working https://github.com/dragonflydb/dragonfly-operator/pull/201

alexanderccc:
@Pothulapati do you need the commit back into the main branch? Assuming it's failing there, I think this might be fixed in 2 ways

Keep the chart and app separate, and have another workflow for helm, this also means having separate versions for the chart which might increase the complexity a bit as there would be 2 different artifacts that need to be maintained and released as part of the same repo with different versions.

Or keep it with matching versions and consider helm as part of the app. Any changes to helm are released with a new version of the app as well (common helm headache) and keep the workflow with sed for version change, login, package and push. For local install (from the cloned repo) there could be a note in the readme stating that the chart and app version need to be updated.

If Helm gets out of hand (which it might), this can be separated into its own repo where you only have the chart.

Aym3nTN:
The fix is easy. Simply add`fetch-depth: 0` to `actions/checkout@v3` step and `git push` in the step where you update Chart.yaml file with the tag version. 
But the concern here is that the Helm Chart version and Operator version are different, the former is at v1.x.x > and the former at v0.1.x. Not the same, I'd suggest working on improving the version management of both components. I'm opening another issue for this initiative.

Issue #141, opened 2024-01-07
Support TLS on admin port


tekumara:
To encrypt traffic between the operator and the dragonfly instances.

PR #176, opened 2024-04-09
feat: enable tls replication on admin port when tls certs are given

nujragan93:
fixes: https://github.com/dragonflydb/dragonfly-operator/issues/141


linouxis9:
Hi! Any updates on this subject? @Pothulapati @Abhra303

PR #187, opened 2024-05-19
feat(service): add support for custom service name

### LINKED ISSUE #178: Ability to configure service name
It would be awesome if we could define the name of the service. Currently the service name is created statically by the name defined in the `Dragonfly` CR.  Being able to customize the service name could be very helpful and speed up a migration of an existing Redis service.

If accepted it would make sense to add it to the `serviceSpec` field.

serviceSpec:
    name: dragonfly-us-west-1

Abhra303:
closes #178


Pothulapati (approved):
LGTM! ⭐

PR #190, opened 2024-05-28
feat: initial helm chart

onedr0p:
This is taking over https://github.com/dragonflydb/dragonfly-operator/pull/174 with commits preserved.


onedr0p:
There's quite a bit of CI changes needed and it would. be best if a maintainer could handle the CI pieces.

PR #198, opened 2024-06-10
fix(rbac-proxy): use the correct version

Abhra303:
The image that we use for rbac-proxy doesn't have 0.17.1 version. The kubebuilder proxy image has the latest `v0.16` version. As a result, the proxy container is not starting. Use the correct latest version.


Pothulapati (approved):

jurim76:
Any plans to upgrade helm chart also? Seems like helm chart version is not updated

PR #203, opened 2024-06-17
fix(rolling-update): avoid replica takeover for single pod instance

### LINKED ISSUE #196: Rollout stuck when there is only a single pod
I am encountering an issue in our dev cluster where we are running dragonfly with a single pod only (as in "`spec.replicas` is set to 1"). When the pod spec changes, the operator is unable to update the stateful set. It will produce these logs:

```
INFO	Rolling out new version
INFO	0/0 replicas are in stable state
ERROR	could not get latest replica
```

From what I understand, the issue comes from [this code in the reconciliation loop](https://github.com/dragonflydb/dragonfly-operator/blob/179803b9a9bf1f0cc92f085250659194ffab8491/internal/controller/dragonfly_controller.go#L189-L193). It seems to assume that there is at least one replica that can eventually take over as master, which is not the case when I had only one pod to begin with.

Abhra303:
Closes #196 and remove unused code.

### DIFF
--- a/internal/controller/dragonfly_controller.go
+++ b/internal/controller/dragonfly_controller.go
@@ -138,7 +138,7 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		fullSyncedUpdatedReplicas := 0
 		for _, replica := range replicas {
 			// Check only with latest replicas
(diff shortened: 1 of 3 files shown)


Pothulapati (approved):
LGTM

PR #222, opened 2024-07-23
fix(reconciler): recreate missing resources

### LINKED ISSUE #138: controller stuck when rollout fails
When pushing a change on a DragonFly resource, a rollout, if an other update is pushed, the controller will wait until first rollout is done before applying any change.

This is an issue, as if the first change contains a typo (invalid image url for instance), there is no way to fix it, as the change with the right image url will never be applied.

To reproduce: 
* deploy an operator and a DragonFly instance with 2 replicas.
* update the dragonfly instance and specify an image with an non existing tag.
* push an other change with a good tag.

The controller wait for the first change to be fully applied but it never occurs as the pod are failing to start with ImagePullError.

### LINKED ISSUE #165: Statefulset update / recreate
Due to config adjustments, the operator tries to patch the sts in a way that would be incompatible.

I am getting this error then:
```
StatefulSet.apps "drangonfly-app" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

It would be great if drangonfly-operator could delete the sts with cascade=orphan and then re-create it with the current config to ensure the desired state.

I was thinking that when i now remove this sts manually, the operator would re-create it to ensure the desired state.
This was also not the case and i would like the operator to ensure the desired state of having the sts with the desired configuration re-created as well.

Abhra303:
Fixes #138
Fixes #165

### DIFF
--- a/internal/controller/dragonfly_controller.go
+++ b/internal/controller/dragonfly_controller.go
@@ -346,47 +346,6 @@ func (r *DragonflyReconciler) getMissingResources(ctx context.Context, df *dfv1a
 	return missingResources, nil
 }
 
-// func (r *DragonflyReconciler) isPVCSpecChanged(df *dfv1alpha1.Dragonfly, sts *appsv1.StatefulSet) bool {
-// 	dfPVCSpec := getPVCSpecFromDragonfly(df)
-// 	stsPVCSpec := getPVCSpecFromStatefulSet(sts)
-
-// 	return !isPVCSpecEqual(dfPVCSpec, stsPVCSpec)
-// }
-
-// func getPVCSpecFromDragonfly(df *dfv1alpha1.Dragonfly) *corev1.PersistentVolumeClaimSpec {
-// 	if df.Spec.Snapshot != nil {
-// 		return df.Spec.Snapshot.PersistentVolumeClaimSpec
-// 	}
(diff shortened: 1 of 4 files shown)


internal/controller/dragonfly_controller.go
```diff
@@ -234,63 +285,115 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		if errors.IsNotFound(err) {
+			missingResources = append(missingResources, resource)
+		} else if err != nil {
+			return nil, fmt.Errorf("failed to get resource %s/%s: %w", df.Namespace, resource.GetName(), err)
+		}
+	}
+	return missingResources, nil
 }
 
+// func (r *DragonflyReconciler) isPVCSpecChanged(df *dfv1alpha1.Dragonfly, sts *appsv1.StatefulSet) bool {
```
Pothulapati: Is this Code still needed? 🤔

Abhra303: removed

Pothulapati (approved):

PR #237, opened 2024-08-26
fix(manifests): Separate the crds from the operator manifests

omar-mohamed-khallaf:
Trying to install the operator with flux kustomization pointing to the manifest directory results in error relating to crds being duplicated.

This commit separates the different manifests in different directories so they don't get combined.


Smana:
That is indeed a good idea. I'm using Flux and I want to install the CRDs in a distinct step.

omar-mohamed-khallaf:
I think this might help you:
```yaml
---
apiVersion: source.toolkit.fluxcd.io/v1
kind: GitRepository
metadata:
  name: dragonfly-operator
  namespace: storage-system
spec:
  interval: 168h
  url: https://github.com/dragonflydb/dragonfly-operator.git
  ref:
    branch: main
  ignore: |
    # exclude all
    /*
    # include charts directory
    !/charts/
---
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
metadata:
  name: dragonfly-operator
  namespace: storage-system
spec:
  interval: 1h
  chart:
    spec:
      chart: charts/dragonfly-operator
      sourceRef:
        kind: GitRepository
        name: dragonfly-operator
        namespace: storage-system
```

Smana:
thank you @omar-mohamed-khallaf , however you don't really separate the crds from the helm chart here?
Actually I want to use dragonfly as the k/v solution for my repo [here](https://github.com/Smana/cloud-native-ref), and I always deploy the crds first.

omar-mohamed-khallaf:
Yes, the crds are installed with the chart.

If it's necessary you may  use [Post Renderers](https://fluxcd.io/flux/components/helm/helmreleases/#post-renderers) to omit other resources at the step of installing crds.

Then use the chart again to install previously omitted resources at the appropriate step.

PR #246, opened 2024-10-02
fix(manifests): fix invalid yaml in crds.yaml

XLordalX:
crds.yaml contains a lot of invalid yaml. This pr fixes that.


Pothulapati:
@XLordalX These are not invalid, and are automatically generated by kube-builder.

XLordalX:
@Pothulapati 

I mean, any yaml linter will tell you otherwise. Besides, nodejs yaml is unable to parse it (yes, after helm templating) which is problematic when used in combination with tools like cdk8s.

PR #252, opened 2024-10-21
feat(operator): allow user to modify imagePullPolicy

cyril-corbon:
Allow user to modify `imagePullPolicy` and set it to Never or IfNotPresent
I let `Always` as default to not "break" anything on old setup


Abhra303:
Thanks! Looking good, Will approve once tests pass.

Abhra303 (approved):

Issue #231, opened 2024-08-19
Set containerSecurityContext for the dragonfly custom resource


geckiss:
Hello,

Right now the dragonfly crd doesn't seem to support setting any **containerSecurityContext**. Some environments require containers to run under non-root user.
Helm chart already supports this - https://github.com/dragonflydb/dragonfly/blob/1f36c9952dd5537fb004d408ffebc977bc4cce8b/contrib/charts/dragonfly/values.yaml#L42
Would it be possible to add this please? It would be a nice addition together with https://github.com/dragonflydb/dragonfly-operator/issues/229

So we can create something like this:

```
apiVersion: dragonflydb.io/v1alpha1
kind: Dragonfly
metadata:
  labels:
    app.kubernetes.io/name: dragonfly-sample
    app.kubernetes.io/instance: dragonfly
    app.kubernetes.io/part-of: dragonfly-operator
    app.kubernetes.io/managed-by: helm
    app.kubernetes.io/created-by: dragonfly-operator
  name: dragonfly-sample
spec:
  containerSecurityContext:
    runAsUser: 1000
    runAsNonRoot: true
```

PR #255, opened 2024-11-01
feat(operator): Add pod security context and container security context to dragonfly sts's pods 

sigarus:
Add pod security context and container security context to dragonfly stateful set's pods by adding them to dragonfly custom resource. (issue https://github.com/dragonflydb/dragonfly-operator/issues/231)


sigarus:
Added ability of watching defined namespace(s) for dragonfly custom resource.

Abhra303 (approved):
LGTM, Thanks!

Abhra303:
can you drop helm changes please? I will add them before releasing.

PR #264, opened 2024-11-19
fix(operator): Enable support for IPv6 first clusters

### LINKED ISSUE #192: Operator does not support IPv6 first environments
I'm currently attempting to run dragonfly with operator in an IPv6 first k3s cluster (it's a complicated setup, I cluster wide IPv4 first isn't an option, and I can't seem to convince calico to use IPv4 only or IPv4 first). The ability to automatically setup a HA setup without having to worry about separate sentinels is very helpful as I would like a HA backend for services depending on redis but also cannot modify them to use the sentinel system.

It would appear that the use of [`Sprintf in dragonfly_instance.go`](https://github.com/dragonflydb/dragonfly-operator/blob/1f3a8dda1957aa06077c0c7f968849a121c686c9/internal/controller/dragonfly_instance.go#L344) is preventing the operator from connecting to the node with "too many colons in address" as documented in redis/go-redis#2467.

It would appear this would be preferable for either a headless service similar to how [cockroachdb performs discovery](https://github.com/cockroachdb/helm-charts/blob/0ad265c0e2c110dde68453ff0c0a2649916a0ee3/cockroachdb/templates/service.discovery.yaml), or the address should be parsed with [net/JoinHostPort](https://pkg.go.dev/net#JoinHostPort) (probably in a util function because that's getting into DRY territory).

I'm happy to have a go at implementing either of these methods however before I do I wanted to get a bit of feedback about which method would be considered the best for this circumstance. Both implementations would make everything more robust, personally I'm in favour of the simpler `JoinHostPort` method, however it may cause issues with TLS if this is later enabled on the admin port, using DNS would prevent this being an issue, but that could also be premature optimisation.

Given this is currently stopping me deploying HA dragonfly I would like to test one of these methods soon and then submit a PR once I've deployed it internally and confirmed the stability, so early feedback is appreciated :)

tomtom5152:
Replace `Sprintf` with `JoinHostPort` to support colons in IPv6 addresses.

Closes #192.

I've only just had a chance to make the changes and haven't got infrastructure in place to build and push an image outside of the main distribution channel. It compiles and runs locally.

### DIFF
--- a/internal/controller/dragonfly_instance.go
+++ b/internal/controller/dragonfly_instance.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
+	"strconv"
 	"strings"
 
 	dfv1alpha1 "github.com/dragonflydb/dragonfly-operator/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


Abhra303 (approved):
LGTM, Thanks!

PR #269, opened 2024-12-05
chore: try to workaround master label with ipv6 issue

cyrinux:


internal/controller/dragonfly_instance.go
```diff
@@ -168,15 +168,25 @@ func (dfi *DragonflyInstance) masterExists(ctx context.Context) (bool, error) {
...
 	}
 
 	for _, pod := range pods.Items {
-		if pod.Status.Phase == corev1.PodRunning && pod.Status.ContainerStatuses[0].Ready && pod.Labels[resources.Role] == resources.Master {
-			return pod.Status.PodIP, nil
+		if pod.Status.Phase == corev1.PodRunning &&
+			pod.Status.ContainerStatuses[0].Ready &&
+			pod.Labels[resources.Role] == resources.Master {
+
+			masterIp, hasMasterIp := pod.Annotations[resources.MasterIp]
```
Pothulapati: If its a master, we don't expect the annotation to be set here right? I'm not sure how this specific change is useful 🤔

Pothulapati (commented):
The masterIP label is used to match if the dragonfly replica state matches with what was configured by the Operator. There should not be any issue in moving this to an annotation (and it makes sense).

The compatibility story is important for sure, My preference would be to do them both for now i.e set as label (if ipv4 only) and annotation (for both ipv4 and ipv6) for a version and remove the label part permanently in future versions. Wdyt?

Also, You should be setting/and using the masterIP here at https://github.com/dragonflydb/dragonfly-operator/blob/2afc70fe7c24e9285bc57521e46dfdded4cd22a6/internal/controller/dragonfly_instance.go#L248

internal/controller/dragonfly_instance.go
cyrinux: Yes, i'm agree, I spot this too, I maybe broke it during the fix/my several refacto !

cyrinux:
> Thanks for all the work!
> 
> The masterIP label is used to match if the dragonfly replica state matches with what was configured by the Operator. There should not be any issue in moving this to an annotation (and it makes sense).
> 
> The compatibility story is important for sure, My preference would be to do them both for now i.e set as label (if ipv4 only) and annotation (for both ipv4 and ipv6) for a version and remove the label part permanently in future versions. Wdyt?
> 
> Also, You should be setting/and using the masterIP here at
> 
> https://github.com/dragonflydb/dragonfly-operator/blob/2afc70fe7c24e9285bc57521e46dfdded4cd22a6/internal/controller/dragonfly_instance.go#L248

This is OK for me, I will do this :-)

cyrinux:
I'm ready for another review @Pothulapati 😃

cyrinux:
I add another commit to bind by default on dualstack, this should not hurt.

cyrinux:
@Pothulapati may I be worring about the CI state ? Can you please re-run it for me or have idea about what happen ? Or all is good ? 😄

cyrinux:
Hi @Pothulapati  can you help me with this PR please ? 🙏🏻

cyrinux:
Hi, anybody can help me passing this PR ? Its running without issue in my prod since weeks.

internal/controller/dragonfly_instance.go
```diff
@@ -322,11 +329,20 @@ func (dfi *DragonflyInstance) replicaOf(ctx context.Context, pod *corev1.Pod, ma
...
+	pod.Annotations[resources.MasterIp] = masterIp
+
+	// for compatibily, to be remove in futur version
+	if !strings.Contains(masterIp, ":") {
+		pod.Labels[resources.MasterIpLabel] = masterIp
+	}
+
 	if err := dfi.client.Update(ctx, pod); err != nil {
-		return fmt.Errorf("could not update replica label")
+		return fmt.Errorf("could not update replica metadatas: %w", err)
```
Ais8Ooz8: The noun metadata is uncountable. The plural form of metadata is also metadata.

Abhra303:
I will check

cyrinux:
Hi here,

@Abhra303 @Pothulapati can I somehow help to get this PR merged ? Do you need something (but time) ?

cyrinux:
Hi,
To let you know I just finally rebase, and I use it, with dragonfly 1.32 on ipv6. All good. 
Can you help me getting merged this one ? 
Thank you

Abhra303:
Hi, sorry for not responding to the PR. Taking a look now.

dhess:
Hi, what's the status on this PR? It's over 9 months old now and the operator still doesn't properly support IPv6.

cyrinux:
Hi, I just rebase my PR, anybody can help get this merged ?

cyrinux:
Hi, I rebase and fix the CI, anybody can help get it merged please ? I see more and more people using ipv6, would be cool.

dhess:
*Please* can we get this merged. This PR is over a year old!

internal/controller/util.go
```diff
@@ -173,3 +174,23 @@ func isMasterError(err error) bool {
 		errors.Is(err, ErrNoHealthyMaster) ||
 		errors.Is(err, ErrIncorrectMasters)
 }
+
+// Helper function to parse Redis INFO data
+func parseRedisInfo(info string) map[string]string {
```
Abhra303: This is unused.

internal/resources/const.go
```diff
@@ -84,9 +84,10 @@ const (
 	KubernetesPartOfLabelKey = "app.kubernetes.io/part-of"
 	KubernetesPartOf         = "dragonfly"
 
-	DragonflyNameLabelKey = "app"
+	MasterIpLabel         string = "master-ip"
```
Abhra303: Why rename it? The existing one is self explanatory enough I believe.

internal/controller/dragonfly_instance.go
```diff
@@ -854,12 +908,16 @@ func (dfi *DragonflyInstance) replTakeover(ctx context.Context, newMaster *corev
 		return fmt.Errorf("response of `REPLTAKEOVER` on replica is not OK: %s", resp)
 	}
 
-	patchFrom := client.MergeFrom(newMaster.DeepCopy())
+	masterIp := newMaster.Status.PodIP
+
 	newMaster.Labels[resources.RoleLabelKey] = resources.Master
-	delete(newMaster.Labels, resources.MasterIpLabelKey)
```
Abhra303, line 860: we shouldn't change this behaviour. The label should work as before. Or atleast update the label to have the new masterIP.

internal/controller/dragonfly_instance.go
```diff
@@ -338,6 +344,33 @@ func (dfi *DragonflyInstance) getMaster(ctx context.Context) (*corev1.Pod, error
 	return &healthyMasters[0], nil
 }
 
+// getMasterIp gets the IP of the master pod, checking annotations first for IPv6 support
+func (dfi *DragonflyInstance) getMasterIp(ctx context.Context) (string, error) {
```
Abhra303: its unused. Remove.

internal/controller/dragonfly_instance.go
```diff
@@ -448,12 +484,20 @@ func (dfi *DragonflyInstance) replicaOf(ctx context.Context, pod *corev1.Pod, ma
...
-	pod.Labels[resources.MasterIpLabelKey] = masterIp
-	if err := dfi.client.Patch(ctx, pod, patchFrom); err != nil {
-		return fmt.Errorf("failed to update the role label on the pod: %w", err)
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+	pod.Annotations[resources.MasterIp] = masterIp
+
+	// for compatibily, to be remove in futur version
+	if !strings.Contains(masterIp, ":") {
```
Abhra303: add this label irrespective of its ip type.

internal/controller/dragonfly_instance.go
```diff
@@ -487,12 +531,22 @@ func (dfi *DragonflyInstance) replicaOfNoOne(ctx context.Context, pod *corev1.Po
...
+	masterIp := pod.Status.PodIP
+
+	dfi.log.Info("Marking pod role as master", "pod", pod.Name, "masterIp", masterIp)
 	pod.Labels[resources.RoleLabelKey] = resources.Master
-	delete(pod.Labels, resources.MasterIpLabelKey)
-	if err := dfi.client.Patch(ctx, pod, patchFrom); err != nil {
-		return fmt.Errorf("failed to update the role label on the pod: %w", err)
+	// for compatibily, to be remove in futur version
+	if !strings.Contains(masterIp, ":") {
+		pod.Labels[resources.MasterIpLabel] = masterIp
```
Abhra303: please add the label irrespective of its IP type. We will remove it in future. Until then the label should exist. Also you are resetting masterIP here. But in replTakeover, you are not setting masterIP. Please maintain same behaviour across.

internal/controller/dragonfly_instance.go
cyrinux: If  I remember its the main point, the label max lenght is too short to be able to store an ipv6.

cyrinux:
Hi @Abhra303, thanks for the review, can you give another eyes, have done the cleanup you told me. 🙏🏻

internal/resources/const.go
```diff
@@ -84,9 +84,10 @@ const (
...
+	DragonflyNameLabelKey        = "app"
 
-	MasterIpLabelKey = "master-ip"
+	MasterIp = "operator.dragonflydb.io/masterIP"
```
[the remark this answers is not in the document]
Abhra303: @cyrinux I agree, you should rename it to MasterIpAnnotationKey.

internal/resources/const.go
Abhra303: @cyrinux You didn't rename it back to MasterIpLabelKey. Please do.

internal/resources/const.go
```diff
@@ -99,6 +100,8 @@ var DefaultDragonflyArgs = []string{
 	"--alsologtostderr",
 	"--break_replication_on_master_restart=true",
 	"--primary_port_http_enabled=false",
+	"--bind=::",
+	"--admin_bind=::",
```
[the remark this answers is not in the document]
Abhra303: @cyrinux why do we need to explicitly set these flags? the default values should work right?

internal/controller/dragonfly_instance.go
Abhra303: ah, ok. But why remove this line? we need to keep the behaviour same for ipv4 only setup. So please delete the MasterIpLabelKey here. This would not have any impact for ipv6 anyway.

internal/controller/dragonfly_instance.go
```diff
@@ -854,12 +878,19 @@ func (dfi *DragonflyInstance) replTakeover(ctx context.Context, newMaster *corev
 		return fmt.Errorf("response of `REPLTAKEOVER` on replica is not OK: %s", resp)
 	}
 
-	patchFrom := client.MergeFrom(newMaster.DeepCopy())
+	masterIp := newMaster.Status.PodIP
+
 	newMaster.Labels[resources.RoleLabelKey] = resources.Master
-	delete(newMaster.Labels, resources.MasterIpLabelKey)
+	// for compatibily, to be remove in futur version
+	newMaster.Labels[resources.MasterIpLabel] = masterIp
```
Abhra303: Why? Remove this line. We don't need to set masterIp label here. We delete the label for masters in ipv4. So for compatiblity, we should delete the label and not reset it.

internal/controller/dragonfly_instance.go
```diff
@@ -487,12 +501,22 @@ func (dfi *DragonflyInstance) replicaOfNoOne(ctx context.Context, pod *corev1.Po
 		}
 	}
 
-	dfi.log.Info("marking pod role as master", "pod", pod.Name)
-	patchFrom := client.MergeFrom(pod.DeepCopy())
+	masterIp := pod.Status.PodIP
+
+	dfi.log.Info("Marking pod role as master", "pod", pod.Name, "masterIp", masterIp)
 	pod.Labels[resources.RoleLabelKey] = resources.Master
-	delete(pod.Labels, resources.MasterIpLabelKey)
```
Abhra303, line 494: add back this line. We delete the label for master. It doesn't impact ipv6.

internal/controller/dragonfly_instance.go
```diff
@@ -487,12 +501,22 @@ func (dfi *DragonflyInstance) replicaOfNoOne(ctx context.Context, pod *corev1.Po
...
-	if err := dfi.client.Patch(ctx, pod, patchFrom); err != nil {
-		return fmt.Errorf("failed to update the role label on the pod: %w", err)
+	// for compatibily, to be remove in futur version
+	if !strings.Contains(masterIp, ":") {
+		pod.Labels[resources.MasterIpLabel] = masterIp
```
Abhra303: We don't need to set the label for ipv4 as the current behaviour is to delete the label for master.

cyrinux:
@Abhra303 after doing the review 2 patch, and build then test, I can see that we break all, I mean, as I said, the main issue I fix with this PR was the fact that an ipv6 can enter in a label, for ex, I now get this error:

```
dragonfly-operator-controller-manager-d7cddfb84-vzkj7 manager 2025-12-16T17:35:07Z	ERROR	Reconciler error	{"controller": "DragonflyPodLifecycle", "controllerGroup": "", "controllerKind": "Pod", "Pod": {"name":"dragonfly-hiboo-1","namespace":"hiboo"}, "namespace": "hiboo", "name": "dragonfly-hiboo-1", "reconcileID": "92a0ef49-1404-415a-8fb3-590fa7b730a9", "error": "failed to configure pod as replica: could not update replica metadata: Pod \"dragonfly-hiboo-1\" is invalid: metadata.labels: Invalid value: \"fd58:312f:6f35:231:6757:ed1:8c2c:3133\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')"}
```

dhess:
Yes, labels for IPv6-first/IPv6-only clusters are precisely the problem, and the entire reason for this PR in the first place.

cyrinux:
I re-add a commit to just add the label if its an ipv4 and will re-test, I confess since the time the algo was not fresh in my head, rebase was not easy too, hard to come back on this, but would be happy if we can finish this one, thanks for your help.

cyrinux:
@Abhra303 after applying your review and a patch for my last comment, the actual state(my lasts commits) works for me on my ipv6 cluster.

Regarding the parameters that need to be add to bind on ipv6 I add them in the documentation, but, to let you know, this shouldn't be an issue to add them by default as binding on ipv6 mean binding on ipv4 too.
I did this test adding the bind="::" and admin_bind="::" on my ipv4 cluster, and it works well, it listen on v6 and v4 but use ipv4 ips.

I don't want to make this a blocker so we can keep this in the documentation.

cyrinux:
I will be happy if somebody else can test to be sure, but I try on a ipv4 cluster too, its ok.

Abhra303:
can you please sign your commits with `-s` flag?

cyrinux:
I will squash, rebase and sign thanks

cyrinux:
It's good for me, done.

Abhra303:
Thanks!

Issue #256, opened 2024-11-03
Grafana Dashboard in Helm Chart doesn't work


maxpain:
Hello. I've deployed the dragonfly-operator Helm chart with the following values:

```yaml
serviceMonitor:
  enabled: true

grafanaDashboard:
  enabled: true
```

But getting this error:

<img width="1876" alt="image" src="https://github.com/user-attachments/assets/c238ceec-7f2c-4485-9d8b-28a9e8e7c8e7">

Pothulapati:
@maxpain With Dragonfly, We don't deploy the prometheus. The issue seems to be that the datasource is misconfigured. Can you check for the same?

Thanks

maxpain:
Yes, the provisioned dashboard doesn't have datasource variable

Pothulapati:
You should be provisioning a prometheus instance right? if you already have one, add it into grafana and set it as the datasource of the dashboard.

maxpain:
@Pothulapati Yes, I have configured Prometheus (VictoriaMetrics, which is much better) for my Grafana installation.
All my Grafana dashboards are declaratively provisioned via ConfigMaps. Almost all of them are from Helm charts and are working fine (CNPG, for example).

So, the Dragonfly grafana dashboard from the helm chart should be fixed.

Pothulapati:
@maxpain Sure!

So, probably a way to pass the prometheus data source (including VictoriaMetrics) as a config through the helm chart?

maxpain:
@Pothulapati I don't know how exactly, but all other helm charts with built-in grafana dashboards don't require configuring the Prometheus data source via helm values. They work out of the box.

Pothulapati:
Looks like DS_Prometheus is a good answer where the user gets the option to select the datasource at the top.

risen228:
I also encountered a similar problem - no datasource selection and the entire dashboard is empty with errors. I was able to fix it locally by adding [datasource selection similar to cloudnative-pg](https://github.com/cloudnative-pg/grafana-dashboards/blob/main/charts/cluster/grafana-dashboard.json#L9057-L9075) to `templating.list`:

```diff
- {
-   "hide": 2,
-   "name": "DS_PROMETHEUS",
-   "query": "prometheus",
-   "skipUrlSync": true,
-   "type": "constant"
- },
+ {
+   "current": {
+     "selected": false,
+     "text": "Prometheus",
+     "value": "prometheus"
+   },
+   "hide": 0,
[...]

PR #279, opened 2024-12-23
fix(monitoring): remove datasource hardcode (#256)

risen228:
Described here: https://github.com/dragonflydb/dragonfly-operator/issues/256#issuecomment-2556772429

These changes will allow grafana to select datasource of type prometheus automatically


risen228:
The formatting is broken due to mixing tabs and spaces =( But overall the whole file is crooked, I would suggest formatting according to the standard, but that's up to you

Abhra303 (approved):
Thanks!

PR #299, opened 2025-03-04
fix(helm): truncate metrics service name to comply with DNS naming spec

### LINKED ISSUE #298: Helm Chart service names can be too long
Take the helm chart service name as an example:

```
name: {{ include "dragonfly-operator.fullname" . }}-controller-manager-metrics-service
```

but the fullname template truncates at 63 chars, which does not include the "-controller-manager-metrics-service" suffix. This can lead to a service name that is too long and, which will cause the resource to not be created.

The same is true for the ClusterRoleBinding, ClusterRole, RoleBinding and Role resources, but AFAIK these are not limited by the DNS spec.

nwp-mai:
This pull request includes changes to the Helm chart to ensure compliance with DNS naming specifications.

* Added a new template definition for `metricsServiceName` to truncate the fullname template to 28 characters and append a suffix, ensuring the total length does not exceed 63 characters. (`charts/dragonfly-operator/templates/_helpers.tpl`)
* Updated the service name in `service.yaml` to use the new `metricsServiceName` template. (`charts/dragonfly-operator/templates/service.yaml`)

fixes #298

### DIFF
--- a/charts/dragonfly-operator/templates/service.yaml
+++ b/charts/dragonfly-operator/templates/service.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ include "dragonfly-operator.fullname" . }}-controller-manager-metrics-service
+  name: {{ include "dragonfly-operator.metricsServiceName" . }}
   labels:
     {{- include "dragonfly-operator.labels" . | nindent 4 }}
     app.kubernetes.io/component: kube-rbac-proxy
--- a/charts/dragonfly-operator/templates/service.yaml
+++ b/charts/dragonfly-operator/templates/service.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ include "dragonfly-operator.metricsServiceName" . }}
+  name: {{ include "dragonfly-operator.controllerServiceName" . }}
   labels:
     {{- include "dragonfly-operator.labels" . | nindent 4 }}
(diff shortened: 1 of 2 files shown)


charts/dragonfly-operator/templates/service.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ include "dragonfly-operator.fullname" . }}-controller-manager-metrics-service
+  name: {{ include "dragonfly-operator.metricsServiceName" . }}
```
Pothulapati: Instead of truncating it to the first 28 characters, I'd refer simplifying the name alltogether.

How about just `fullname-controller-service` or just `fullname-controller-svc`?

Pothulapati (changes requested):

charts/dragonfly-operator/templates/service.yaml
matt-mai: Sure, we can do that, but the fullname will still get generated from the release name the user chooses, so there's always a chance that it will be too long and needs to be truncated. A shorter suffix won't change that, just make it less likely.

I chose to truncate the fullname to keep the suffix intact, but we could also just truncate at the end.

Unless you want to define a hardcoded name? That would cause issues if a user decides to have two helm releases in the same namespace for some reason

Pothulapati: maybe truncate what the user passes to a specific size and then have the controller-svc suffix?

Pothulapati (approved):

PR #308, opened 2025-03-19
fix(pod-lifecycle): improve replication management

Abhra303:
Replication role for replicas is actually "slave" and not "replica".


internal/controller/dragonfly_instance.go
```diff
@@ -223,7 +223,7 @@ func (dfi *DragonflyInstance) checkReplicaRole(ctx context.Context, pod *corev1.
 		}
 	}
 
-	if redisRole != resources.Replica {
+	if redisRole == resources.Master {
```
romange, line 226: it's a good fix.

romange (approved):

PR #310, opened 2025-03-25
fix: improve reconciling dragonfly resource

ihor-hrytskiv:
Improve the reconciliation of Dragonfly resources; avoid excessive looping.

### DIFF
--- a/internal/controller/util.go
+++ b/internal/controller/util.go
@@ -34,8 +34,8 @@ import (
 )
 
 const (
-	PhaseResourcesCreated string = "ResourcesCreated"
-	PhaseReady            string = "Ready"
+	PhaseResourcesCreated string = "resources-created"
+	PhaseReady            string = "ready"
 )
 
 // isPodOnLatestVersion returns if the Given pod is on the updatedRevision


ihor-hrytskiv:
Hi @Abhra303, Could you please take a look at one more pull request?

Abhra303:
Taking a look today, Thanks for the PR!

internal/controller/util.go
```diff
@@ -27,13 +27,15 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
 const (
-	PhaseResourcesCreated string = "resources-created"
-
-	PhaseReady string = "ready"
+	PhaseResourcesCreated string = "ResourcesCreated"
+	PhaseReady            string = "Ready"
```
Abhra303: It will break backward-compatibility. The new phase names looks good. But You need to make sure the existing deployments with old phase statuses won't get affected. The new operator version won't recognize the old phases. Please open a separate PR for this with backward-compatible changes.

ihor-hrytskiv: Thanks @Abhra303 , you're right, I've reverted changes and will open new PR for this.

PR #318, opened 2025-04-25
feat: allow to pass extra args to containers

emaincourt:
**Title:**  Allow to pass extra flags to containers

**Description:** This PR allows to pass extra args to both the rbac proxy and the manager. I mostly need this change in order to configure the logging level/encoding.

### DIFF
--- a/charts/dragonfly-operator/templates/deployment.yaml
+++ b/charts/dragonfly-operator/templates/deployment.yaml
@@ -43,6 +43,9 @@ spec:
             - --upstream=http://127.0.0.1:8080/
             - --logtostderr=true
             - --v=0
+            {{- with .Values.rbacProxy.extraArgs}}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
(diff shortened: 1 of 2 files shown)


charts/dragonfly-operator/templates/deployment.yaml
```diff
@@ -43,6 +43,9 @@ spec:
             - --upstream=http://127.0.0.1:8080/
             - --logtostderr=true
             - --v=0
+            {{- with .Values.rbacProxy.extraArgs}}
```
Pothulapati, line 46: Just trying to understand why would you want to pass more args to the controller and the rbac proxy itself? (I can see the reason for Dragonfly args and we already have it.)

emaincourt: Hi,

For the manager, this change would allow to set flags such as `-zap-log-level` and `-zap-encoder` for instance, which is not currently feasible if I'm right. For the rbac proxy there are many available flags that could be tuned by any user I assume.

Pothulapati (approved):

PR #345, opened 2025-08-15
feat: additional containers and volumes support

iyuroch:
This PR's introduces new feature which allows operator user to add containers and volumes to statefulset. User can optionally override container definition of dragonfly container itself.
We are using this feature to add sidecar container for snapshot cleanup, but one can use it as logging sidecar or any other sidecar containers user want to use.


iyuroch:
@Abhra303 could you please take a look at this pr - I'm adding e2e test right now, could you check if you want me to add/change something

Issue #340, opened 2025-08-08
Expose the ability to configure persistentVolumeClaimRetentionPolicy


robinsla-deshaw:
The DragonflyDB operator leaves around the PVCs when a user deletes the Dragonfly resource. This is because when it deletes the StatefulSet, Kubernetes leaves the PVCs as the persistentVolumeClaimRetentionPolicy field of the StatefulSet defaults to Retain

Can you please expose the ability to set this to Delete on both `whenDeleted` and `whenScaled` ? The default can still be Retain but it would be nice to be able to set it to avoid having to write logic to clean up PVCs.

This functionality used to be behind a feature gate, but it's in GA as of 1.32 so no reason not to expose/support it, its not going anywhere.

Abhra303:
Sure, makes sense. Feel free to contribute.

robinsla-deshaw:
@Abhra303 I've logged #353 to implement it, do let me know how I can move that forward

PR #353, opened 2025-08-20
feat(operator): Add support for PVC retention policy

robinsla-deshaw:
Implements #340 

 * Extended the Dragonfly resource with a `persistentVolumeClaimRetentionPolicy` field using the struct from appsv1
 * Added the conditional in the primary `GenerateDragonflyResource` function to set it in the StatefulSet if present

I've added tests but I've not used this testing harness before so sorry if I've not done it right/put it in the wrong place.

I've tested locally and things work as I'd expect:
 * If the policy isn't set it defaults to Retain
 * If set, the STS reflects it

This would only work on Kubernetes clusters running 1.27+ where StatefulSetAutoDeletePVC went into BETA (GA as of 1.32). That feels like a good enough number of versions/time that we don't need to lock this behind a feature gate/flag, but let me know if you'd want me to add that so this doesn't break on older clusters.

### DIFF
--- a/api/v1alpha1/dragonfly_types.go
+++ b/api/v1alpha1/dragonfly_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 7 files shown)


Abhra303:
Hi @robinsla-deshaw, e2e tests are failing, can you please fix those first?

Abhra303:
Closing as it looks like the PR is not active.

PR #362, opened 2025-09-04
feat[resources]: push custom labels on STS itself

Pinimo:
For compliance, it is useful to be able to customize the labels attached to the STS created for any given Dragonfly instance. Till now, it was possible to customize the labels of pods of the STS, but not labels of the STS itself.

I would need help, please, to add tests about this feature. I have searched around in the test code, but I fail (not the tests! they succeed) to understand how to test for a custom label being added to the STS, as well as for label change reconciliation.

I have "manually tested" and in case of custom label change in the values, the STS will be updated, and the pods rolled out (since they too will have to update the label).

### DIFF
--- a/api/v1alpha1/dragonfly_types.go
+++ b/api/v1alpha1/dragonfly_types.go
@@ -62,7 +62,7 @@ type DragonflySpec struct {
 	// +kubebuilder:validation:Optional
 	Annotations map[string]string `json:"annotations,omitempty"`
 
-	// (Optional) Labels to add to the Dragonfly pods.
+	// (Optional) Labels to add to the Dragonfly pods and StatefulSet.
 	// +optional
 	// +kubebuilder:validation:Optional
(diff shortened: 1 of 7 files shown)


Pinimo:
I prefer the idea in #363 :wink: I'll close (or keep) this PR once a project core contributor has given his or her feeling on the best way to go :slightly_smiling_face:

Abhra303:
Hi  @Pinimo, thank you for working on this issue. Honestly I feel like we should have separate labels and annotations fields for sts and pods. E.g. something like we do in serviceSpec. This way it will be more clear for users and they will be able to control labels for specific resources.
```yaml
stsSpec:
  labels:
  annotations:
podSpec:
  labels:
  annotations:
```

Abhra303:
This feature is added in the recent version. Closing this PR. Feel free to open PRs on other issues!

PR #363, opened 2025-09-09
feat(crd): allow child resources to get additional labels and annotations

gdsoumya:
This PR adds an optional flag in the CRD to pass down the metadata(labels and annotations) of the CR to the child resources. This allows for any custom tracking labels/annots to be passed down to the child resources

### DIFF
--- a/internal/resources/resources.go
+++ b/internal/resources/resources.go
@@ -496,9 +496,11 @@ func generateResourceLabels(df *resourcesv1.Dragonfly) map[string]string {
 		DragonflyNameLabelKey:          df.Name,
 	}
 
-	for key, value := range df.Spec.OwnedObjectsMetadata.Labels {
-		if _, ok := labels[key]; !ok {
-			labels[key] = value
+	if df.Spec.OwnedObjectsMetadata != nil {
+		for key, value := range df.Spec.OwnedObjectsMetadata.Labels {
+			if _, ok := labels[key]; !ok {
+				labels[key] = value
+			}
 		}
 	}
 
@@ -507,8 +509,10 @@ func generateResourceLabels(df *resourcesv1.Dragonfly) map[string]string {
 
 func generateResourceAnnotations(df *resourcesv1.Dragonfly) map[string]string {
 	annotations := map[string]string{}
-	for key, value := range df.Spec.OwnedObjectsMetadata.Annotations {
-		annotations[key] = value
+	if df.Spec.OwnedObjectsMetadata != nil {
+		for key, value := range df.Spec.OwnedObjectsMetadata.Annotations {
+			annotations[key] = value
+		}
 	}
 
 	return annotations


gdsoumya:
@Abhra303 opened a PR directly instead of creating an issue. I know there's another PR https://github.com/dragonflydb/dragonfly-operator/pull/362 that adds something similar but it inherits the values to both the pod and sts. In our usecase we just want the tracking annotations/labels to be passed down to the main child resources managed by the CRD and not to the pods itself.

Pinimo:
Hi! I personally like this PR's idea better than #362 (mine :sweat_smile:). However, it looks like it is still lacking some tests. If you need, you can copy/adapt them from my PR :wink:

gdsoumya:
Thanks for the pointers @Pinimo ! Will add some tests!

config/crd/bases/dragonflydb.io_dragonflies.yaml
```diff
@@ -4606,6 +4606,10 @@ spec:
                   type: object
                   x-kubernetes-map-type: atomic
                 type: array
+              inheritMetadata:
+                description: (Optional) Inherit labels and annotations from Dragonfly
+                  CR to Dragonfly direct child resources like STS, Services, PDB etc.
+                type: boolean
```
Pinimo: Maybe we should make this a default `true`, or even not add the option, to avoid complexifying the configuration. Metadata inheritance sounds like the behaviour I would like to see from an operator in most if not all cases.

gdsoumya: I added a new flag to keep things backwards compat and not suddenly bring a change that could affect others already using the operator in unknown ways.

Maybe after a few releases (probably a major release) it can be turned on permanently

Pinimo: Yes indeed. It also makes sense if we keep the default to false on the first (minor) release. I guess the default value should be documented somewhere, I'm not sure it's the case yet though :wink:

internal/resources/resources.go
```diff
@@ -499,4 +495,24 @@ func generateResourceLabels(df *resourcesv1.Dragonfly) map[string]string {
 		KubernetesManagedByLabelKey:    DragonflyOperatorName,
 		DragonflyNameLabelKey:          df.Name,
 	}
+
+	if df.Spec.InheritMetadata {
+		for key, value := range df.Labels {
```
Abhra303: The labels can cause conflicts with the labels set by controllers. You should ignore the labels that matches our default labels.


```suggestion
		for key, value := range df.Labels {
		   if _, ok := labels[key]; ok {
		     continue
		   }
```

config/crd/bases/dragonflydb.io_dragonflies.yaml
Abhra303: I am not sure about adding a field for this. Why do you think it can break compatibility? If the labels or annotations doesn't override operator given labels (or spec annotations), it shouldn't break compatibility at all.

gdsoumya: If you feel it's okay to just add the change I will remove it. I am not sure how everyone else is using dragonfly and monitoring it so felt it's safer to not introduce the change by default. But seems like it isn't an issue, so removing it.

internal/resources/resources.go
gdsoumya: Updated

config/crd/bases/dragonflydb.io_dragonflies.yaml
gdsoumya: updated

Abhra303:
On second thought, I think you are right to have a field for this. But in general, I prefer a clear and specific field for this. E.g.
```yaml
spec:
  resourcesMetadata:
    labels: ...
    annotations: ...
```

I don't think it is common to pass labels/annotations down to its owned resources directly from CRD's metadata. What do you think?

gdsoumya:
> I don't think it is common to pass labels/annotations down to its owned resources directly from CRD's metadata. What do you think?

I have seen some operators pass down the labels and annotations on the parent resource specially if there's not concept of `template` like in case of deployments/sts etc.

I would have gone for a separate field but i think it'll just confuse people with the existing Spec.labels applied to pods directly. Which is why I had the original field to optionally inherit the labels which makes it an explicit user choice to reuse the parents metadata for all child resources and not just the sts.

Unless we want to introduce breaking change and have a separate spec for pod and sts like we have serviceSpec I would suggest reverting the last commit to have the optional flag to inherit and let users choose the behaviour for all direct child resources and they can specify any specific metadata for pods/service inside the existing spec fields.

Abhra303:
Honestly, the existing `spec.Labels` is confusing anyway (and should be migrated to more pod specific fields instead) because it doesn't say it target pods specifically. I am not in favour of passing down metadata because users then will not able to configure the CRD when they need both cases - (i) unique labels to identify the CRD along with (ii) some common labels for all the resources it owns.

For example:
```yaml
metadata:
  labels:
    operator: "dragonfly"
    crd-group: db
spec:
  ownedObjectsMetadata:
    labels:
      operator: "dragonfly"
    ...
```

Also the reconciler listens to spec changes only. So if you change the metadata of CRD, the reconciler won't be triggered and owned resources will not have the updated metadata labels.

gdsoumya:
Okay then I will go ahead and add this field, 
```
  ownedObjectsMetadata:
    annotations:
       .......
    labels:
       .......
 ```
 
 For the existing spec.Labels I will leave it as is, I don't want to stretch the scope of this PR so maybe a followup to change it to spec.podSpec.labels and deprecating old field will be better

gdsoumya:
@Abhra303 updated

Abhra303:
Thanks!

gdsoumya:
@Abhra303 seems like the CI started failing. it's failing on main too https://github.com/dragonflydb/dragonfly-operator/actions/runs/17944604330/job/51027928621

Don't think the failure is related to my changes. Can you please take a look

Abhra303:
Yea, it is not related to your change. I'll look.

Abhra303:
@gdsoumya, now the test is timing out, can you please check?

gdsoumya:
Pushed fix, hopefully CI will pass now

gdsoumya:
@Abhra303 it passed!

Abhra303:
Thanks!

gdsoumya:
@Abhra303 any eta on a new release with this change?

PR #365, opened 2025-09-15
fix: remove master label before promoting new master

ihor-hrytskiv:
We encountered an issue when a master pod was flapping between `Ready=false` and `Ready=true`.

**This led to the following problem:**
- The operator promoted a replica to master.
- On the next reconciliation, the old master pod was still running and still had the label `role=master`.
- As a result, two pods were labeled as master.
- This triggered the error: incorrect number of masters.

**Fix:**
We now remove the `role=master` label from the old master before promoting a new replica. This ensures only one pod is marked as master at any given time.


Abhra303 (approved):
LGTM

Abhra303:
Can you sign your commit pls?

ihor-hrytskiv:
> Can you sign your commit pls?

@Abhra303 Thanks,  I have signed commit

PR #369, opened 2025-09-18
fix(helm) Fix service monitor path

blake-hamm:
Noticed a path was slightly off. I don't think it has much affect, but I figured I'd provide a fix.


Abhra303:
Hi @blake-hamm, thanks for the PR. pls sign-off your commit.

blake-hamm:
Signed off. Thanks!

PR #396, opened 2025-10-13
feat: add aggregate cluster roles

pujan-rpc:
This PR adds support for aggregated cluster roles


Abhra303:
can you please sign your commit with -s please?

PR #406, opened 2025-10-27
feat: add support for referencing existing PVC for snapshots

### LINKED ISSUE #342: Allow referencing an existing PVC for snapshots
On **GKE** with **Filestore CSI driver**, we’re seeing intermittent provisioning/mount issues when the Dragonfly Operator manages the PVC via `spec.snapshot.persistentVolumeClaimSpec`. In many runs the Operator creates the **PVC**, but the **PV** is not created or the Pod mount phase fails with events claiming the PVC “already exists” — which *is* the very claim the Operator created.  
To unblock production, we’d like the Operator to support **referencing an existing PVC** (pre-created and validated) via something like:

```yaml
spec:
  snapshot:
    existingPersistentVolumeClaimName: <name>
```

Right now the CRD only supports:

```yaml
spec:
  snapshot:
    persistentVolumeClaimSpec: { ... }
```

## Environment

- **Kubernetes (GKE) version:** 1.33.2-gke.1240000
- **Dragonfly Operator version / chart:** dragonfly-operator-v1.1.11 / v1.1.11
- **Dragonfly image version:** docker.dragonflydb.io/dragonflydb/operator:v1.1.11  
- **Filestore CSI driver:** enabled (using GKE-provided storage classes like `standard-rwx` / `enterprise-rwx` as applicable)
- **Network / node pool:** same cluster & node pool used for control tests
- **StorageClass reclaim/binding mode tried:** `WaitForFirstConsumer` (default) and `Immediate` — both hit the same behavior

## What happens

1. Apply a `Dragonfly` CR that defines `spec.snapshot.persistentVolumeClaimSpec` pointing to a Filestore-backed RWX StorageClass.
2. Operator successfully **creates the PVC** (intermittent).
3. Either:
   - No **PV** gets bound (stuck Pending), or
   - Pod mount **fails** with event like “PVC <name> already exists” (while it’s the same, just-created PVC).
4. When the Operator *does* create the claim and the PV binds, the Pod still may fail to mount with the same “already exists” style error.

## What we validated

- **Pre-created PVC works:** If we create a Filestore-backed PVC ourselves and mount it on a **manual** Pod, it binds & mounts fine (no firewall/networking issues). Same cluster, same node pool.
- **Binding mode:** Using a StorageClass with `VolumeBindingMode: Immediate` produces the same symptoms.
- **Cluster/Filestore sanity:** Filestore CSI works for other Pods in the cluster.

## Why this feature is needed

Because the Operator currently **only** accepts an embedded `persistentVolumeClaimSpec`, users cannot point to a **known-good existing PVC** as a workaround. Supporting an `existingPersistentVolumeClaimName` (mutually exclusive with `persistentVolumeClaimSpec`) would:
- Let users pre-provision, validate, and manage the lifecycle of storage independently.
- Unblock production when CSI provisioning is flaky or constrained by platform policies.
- Align with patterns used by other operators that allow both “spec to create” *and* “reference existing” modes.

## Minimal reproducible example (MRE)

**StorageClass (example):**
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: filestore-rwx
provisioner: filestore.csi.storage.gke.io
allowVolumeExpansion: true
parameters:
  tier: ENTERPRISE
  network: default
volumeBindingMode: WaitForFirstConsumer
mountOptions:
  - nconnect=8
```

**Dragonfly CR:**
```yaml
kind: Dragonfly
metadata:
  name: df-filestore
spec:
  replicas: 1
  snapshot:
    # Current API: works intermittently; cannot reference an existing PVC
    persistentVolumeClaimSpec:
      accessModes: ["ReadWriteMany"]
      storageClassName: filestore-rwx
      resources:
        requests:
          storage: 50Gi
```

**Proposed alternative (feature request):**
```yaml
kind: Dragonfly
metadata:
  name: df-filestore
spec:
  replicas: 1
  snapshot:
    # New field (mutually exclusive with persistentVolumeClaimSpec)
    existingPersistentVolumeClaimName: df-snapshots-pvc
```

## Expected behavior

- If `existingPersistentVolumeClaimName` is set, the Operator:
  - Validates the PVC exists and access mode/storage class is compatible.
  - Skips provisioning a new PVC/PV.
  - Mounts the referenced claim into the StatefulSet/Pod used for Dragonfly snapshots.

## Actual behavior

- Operator intermittently creates the PVC but PV doesn’t bind, or Pod mount fails claiming the PVC already exists (even though it’s the same newly created claim).

## Additional context

- Dragonfly Operator docs show only `snapshot.persistentVolumeClaimSpec` today; there’s no documented “use existing PVC” field.
- We confirmed Filestore CSI basics per GKE docs; manual Pods can mount the same StorageClass successfully.

## Proposed API

CRD schema addition (names/validation illustrative):

```yaml
snapshot:
  oneOf:
    - required: ["persistentVolumeClaimSpec"]
    - required: ["existingPersistentVolumeClaimName"]
  properties:
    existingPersistentVolumeClaimName:
      type: string
      description: "Name of an existing PVC to use for Dragonfly snapshots"
```

Controller logic:
- If `existingPersistentVolumeClaimName` → fetch PVC, validate, inject into StatefulSet volume/volumeMounts, skip reconciler path that creates PVC.

rdemoraes:
## Description

This PR adds support for referencing a pre-existing PVC in the snapshot configuration, providing an alternative to having the operator manage PVC creation. This is particularly useful on GKE with Filestore CSI driver where intermittent provisioning/mount issues may occur.

Fixes #342

### API Changes
- Added `existingPersistentVolumeClaimName` field to the `Snapshot` struct in `api/v1alpha1/dragonfly_types.go`
- Field is mutually exclusive with `persistentVolumeClaimSpec`
- Both fields are optional, allowing for S3-based snapshots without PVC

### Controller Logic
- Updated `internal/resources/resources.go` to:
  - Validate mutual exclusivity between `persistentVolumeClaimSpec` and `existingPersistentVolumeClaimName`
  - Use existing PVC via regular Volume (not VolumeClaimTemplate) when `existingPersistentVolumeClaimName` is set
  - Updated cron validation to accept either PVC option

### Documentation & Examples
- Added sample configuration: `config/samples/v1alpha1_dragonfly_existing_pvc.yaml`
- Regenerated CRD manifests with the new field

## How It Works

**Before (creates PVC per pod):**
```yaml
spec:
  snapshot:
    persistentVolumeClaimSpec:
      accessModes: ["ReadWriteMany"]
      storageClassName: filestore-rwx
      resources:
        requests:
          storage: 50Gi
```

**After (uses existing PVC):**
```yaml
spec:
  snapshot:
    existingPersistentVolumeClaimName: df-snapshots-pvc
    cron: "0 */6 * * *"
```

When `existingPersistentVolumeClaimName` is set:
- The operator validates the PVC exists
- Uses the existing PVC via a regular Volume (shared across all pods)
- Skips PVC/PV provisioning logic
- Works best with ReadWriteMany (RWX) storage

## Use Cases

1. **Workaround for CSI Issues**: Pre-create and validate PVC when operator-managed PVCs have intermittent binding issues
2. **Independent Lifecycle**: Manage storage lifecycle separately from the operator
3. **Production Stability**: Use known-good, pre-provisioned storage
4. **Platform Policies**: Comply with org policies requiring manual storage provisioning

## Testing

1. Create a PVC manually:
   ```bash
   kubectl create -f - <<EOF
   apiVersion: v1
   kind: PersistentVolumeClaim
   metadata:
     name: df-snapshots-pvc
   spec:
     accessModes: ["ReadWriteMany"]
     storageClassName: filestore-rwx
     resources:
       requests:
         storage: 50Gi
   EOF
   ```

2. Deploy Dragonfly with existing PVC reference:
   ```bash
   kubectl apply -f config/samples/v1alpha1_dragonfly_existing_pvc.yaml
   ```

3. Verify the Pod uses the existing PVC:
   ```bash
   kubectl describe pod dragonfly-existing-pvc-0
   ```

## Checklist

- [x] Code changes follow the project's style guidelines
- [x] CRD manifests regenerated
- [x] Example configuration provided
- [x] Validation logic ensures mutual exclusivity
- [x] Backward compatible (new field is optional)

### DIFF
--- a/api/v1alpha1/dragonfly_types.go
+++ b/api/v1alpha1/dragonfly_types.go
@@ -237,6 +237,11 @@ type Snapshot struct {
 	// +optional
 	// +kubebuilder:validation:Optional
 	PersistentVolumeClaimSpec *corev1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
+
+	// (Optional) Name of an existing PVC to use for Dragonfly snapshots
+	// +optional
+	// +kubebuilder:validation:Optional
+	ExistingPersistentVolumeClaimName string `json:"existingPersistentVolumeClaimName,omitempty"`
 }
 
 type Authentication struct {
--- a/config/crd/bases/dragonflydb.io_dragonflies.yaml
+++ b/config/crd/bases/dragonflydb.io_dragonflies.yaml
@@ -6404,6 +6404,10 @@ spec:
                   enableOnMasterOnly:
                     description: (Optional) Enable snapshot on master only
                     type: boolean
+                  existingPersistentVolumeClaimName:
+                    description: (Optional) Name of an existing PVC to use for Dragonfly
+                      snapshots
+                    type: string
                   persistentVolumeClaimSpec:
                     description: (Optional) Dragonfly PVC spec
                     properties:
--- a/config/samples/v1alpha1_dragonfly_existing_pvc.yaml
+++ b/config/samples/v1alpha1_dragonfly_existing_pvc.yaml
@@ -1,0 +1,24 @@
+apiVersion: dragonflydb.io/v1alpha1
+kind: Dragonfly
+metadata:
+  labels:
+    app.kubernetes.io/name: dragonfly
+    app.kubernetes.io/instance: dragonfly-existing-pvc
+    app.kubernetes.io/part-of: dragonfly-operator
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/created-by: dragonfly-operator
+  name: dragonfly-existing-pvc
+spec:
+  replicas: 2
+  snapshot:
+    # Use an existing PVC instead of creating a new one
+    existingPersistentVolumeClaimName: dragonfly-snapshots-pvc
+    cron: "0 */6 * * *"  # Snapshot every 6 hours
+  resources:
+    requests:
+      cpu: 500m
+      memory: 500Mi
+    limits:
+      cpu: 600m
+      memory: 750Mi
+
--- a/internal/resources/resources.go
+++ b/internal/resources/resources.go
@@ -231,8 +231,13 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly) ([]client.Object, err
 	}
 
 	if df.Spec.Snapshot != nil {
+		// validate mutual exclusivity of PVC spec and existing PVC name
+		if df.Spec.Snapshot.PersistentVolumeClaimSpec != nil && df.Spec.Snapshot.ExistingPersistentVolumeClaimName != "" {
+			return nil, fmt.Errorf("persistentVolumeClaimSpec and existingPersistentVolumeClaimName are mutually exclusive")
+		}
+
 		// err if pvc is not specified & s3 sir is not present while cron is specified
-		if df.Spec.Snapshot.Cron != "" && df.Spec.Snapshot.PersistentVolumeClaimSpec == nil && df.Spec.Snapshot.Dir == "" {
+		if df.Spec.Snapshot.Cron != "" && df.Spec.Snapshot.PersistentVolumeClaimSpec == nil && df.Spec.Snapshot.ExistingPersistentVolumeClaimName == "" && df.Spec.Snapshot.Dir == "" {
 			return nil, fmt.Errorf("cron specified without a persistent volume claim")
 		}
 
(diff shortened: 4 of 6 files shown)


Abhra303:
@rdemoraes please sign your commit with `-s` flag

rdemoraes:
Thanks @Abhra303 for the heads-up! I’ll make sure to sign off commits going forward.

Issue #283, opened 2025-01-03
Configurable default image


foosinn:
Currently the default image for Dragonfly Instances is hardcoded (see [1]).

For air-gapped setups it would be nice to replace the default value. Maybe configurable via command-line or enviornment?

Thanks!

[1]: https://github.com/dragonflydb/dragonfly-operator/blob/b07f733376e75f4a600345e538c27ff258502e6f/internal/resources/const.go#L36

Pothulapati:
Hey @foosinn,

You can override the image in your `Dragonfly` configuration by setting the image argument accordignly.

 https://github.com/dragonflydb/dragonfly-operator/blob/b07f733376e75f4a600345e538c27ff258502e6f/config/crd/bases/dragonflydb.io_dragonflies.yaml#L1193


Feel free to reopen the issue, if this does not solve your problem.

foosinn:
They @Pothulapati,

thanks for your quick reply.

In our case we would like to set a default while installing the operator. So developers can create their dragonflydb without having to have knowledge about our internal images.


I'm aware that this is potentially out of scope and we also can solve this by wrapping it into a helm chart with some defaults. Please let me know :)

alphabet5:
Could this be re-opened? We have a bunch of different teams and projects using dragonfly instances, and when there is a fix in a specific dragonfly version we either have to wait until the dragonfly operator changes the default image, or override the image in all of the various places. Having this be configurable (even via an env or configmap?) for the operator would make operations a lot better.

Pothulapati:
Ahh, That makes sense actually!

PR #407, opened 2025-10-29
feat: allow configuring the default dragonfly version with an env

alphabet5:
This allows configuring the default dragonfly image with an env passed to the operator. If it's blank or not configured, it will fall back to a static string similar to how it's currently operating.

### DIFF
--- a/internal/resources/version.go
+++ b/internal/resources/version.go
@@ -16,6 +16,13 @@ limitations under the License.
 
 package resources
 
-const (
-	Version = "v1.34.1"
-)
+import "os"
+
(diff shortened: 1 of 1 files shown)


alphabet5:
https://github.com/dragonflydb/dragonfly-operator/pull/421 accomplishes the same thing.

alphabet5:
fixes https://github.com/dragonflydb/dragonfly-operator/issues/283

Abhra303:
Hey, I merged #421 as that one seems a better approach. Thank you for opening this PR. Feel free to fix other issues.

PR #415, opened 2025-11-19
fix(operator): do not reconcile resources if dragonfly object is foreground deleted

### LINKED ISSUE #408: Operator does not support foreground deletion of dragonfly instance
It seems like the operator goes into a loop of destroying and creating objects if a dragonfly instance is deleted with foreground cascade. To repro, create a dragonfly instance (mine is as follows):
```yaml
kind: Dragonfly
metadata:
  name: dragonfly
spec:
  replicas: 2
  snapshot:
    cron: "*/5 * * * *"
    persistentVolumeClaimSpec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
```
If I then do `kubectl delete dragonfly dragonfly` (background cascade), everything is fine. If I do `kubectl delete dragonfly dragonfly --cascade=foreground`, the deletion gets stuck, and if I do `kubectl get pods` or `kubectl get pdb`, I can see the pods / PDB objects are rapidly destroyed and created, like the operator is trying to reconcile the object as it's being deleted. I think the operator should detect if the object is being deleted and prevent reconciliation if so.

We are using the latest version of dragonfly-operator (1.3.0)

rkevin-arch:
Fixes #408

Testing I went through to make sure the issue is resolved:
```bash
minikube start
kubectl apply -f https://raw.githubusercontent.com/dragonflydb/dragonfly-operator/main/manifests/dragonfly-operator.yaml
# wait for opeartor to come up via `kubectl get pods -A`

### BEFORE
kubectl apply -f https://raw.githubusercontent.com/dragonflydb/dragonfly-operator/main/config/samples/v1alpha1_dragonfly.yaml
# wait for pods to be ready via `kubectl get pods -A`
kubectl delete --cascade=foreground -f https://raw.githubusercontent.com/dragonflydb/dragonfly-operator/main/config/samples/v1alpha1_dragonfly.yaml
# verify deletion is getting stuck via `kubectl get pods -A`, `kubectl get sts -A`. you should see one pod creating and terminating. the delete command should also hang forever

# to delete properly 
kubectl delete -f https://raw.githubusercontent.com/dragonflydb/dragonfly-operator/main/config/samples/v1alpha1_dragonfly.yaml

# edit operator image and change the image to the one I just built via `eval $(minikube docker-env) && make docker-build`
kubectl edit deploy -n dragonfly-operator-system dragonfly-operator-controller-manager
# wait for opeartor to come up via `kubectl get pods -A`

### AFTER
kubectl apply -f https://raw.githubusercontent.com/dragonflydb/dragonfly-operator/main/config/samples/v1alpha1_dragonfly.yaml
# wait for pods to be ready via `kubectl get pods -A`
kubectl delete --cascade=foreground -f https://raw.githubusercontent.com/dragonflydb/dragonfly-operator/main/config/samples/v1alpha1_dragonfly.yaml   
# verify deletion went through properly via `kubectl get pods -A`, `kubectl get sts -A`. the delete command should also return quickly

### DIFF
--- a/internal/controller/dragonfly_instance.go
+++ b/internal/controller/dragonfly_instance.go
@@ -384,6 +384,11 @@ func (dfi *DragonflyInstance) patchStatus(ctx context.Context, status dfv1alpha1
 	return nil
 }
 
+// isTerminating returns true if the dragonfly instance is being deleted
+func (dfi *DragonflyInstance) isTerminating() bool {
+	return dfi.df.ObjectMeta.DeletionTimestamp != nil
+}
+
 // detectOldMasters checks whether there are any old masters and deletes them
 func (dfi *DragonflyInstance) detectOldMasters(ctx context.Context, updateRevision string) error {
 	var masterPods corev1.PodList
--- a/internal/controller/dragonfly_controller.go
+++ b/internal/controller/dragonfly_controller.go
@@ -60,6 +60,11 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, client.IgnoreNotFound(fmt.Errorf("failed to get dragonfly instance: %w", err))
 	}
 
+	if dfi.isTerminating() {
+		// Ignore dragonfly instance that is being foreground deleted
+		return ctrl.Result{}, nil
+	}
+
 	if err = dfi.reconcileResources(ctx); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to reconcile dragonfly resources: %w", err)
 	}

### CODE REVIEW

File: internal/controller/dragonfly_instance.go [hunk 1]
```diff
@@ -384,6 +384,11 @@ func (dfi *DragonflyInstance) patchStatus(ctx context.Context, status dfv1alpha1
 	return nil
 }
 
+// isTerminating returns true if the dragonfly instance is being deleted
+func (dfi *DragonflyInstance) isTerminating() bool {
+	return dfi.df.ObjectMeta.DeletionTimestamp != nil
```
moredure, line 389: can be replaced with !dfi.df.ObjectMeta.DeletionTimestamp.IsZero(), up to you, have seen it more frequently in the codebases

rkevin-arch: Looks like this merged as-is so I'm gonna leave it alone. Thanks!

PR #463, opened 2026-02-22
chore(controller): eliminate redundant redis client connections

miledxz:
Adding an update to eliminate redundant Redis client connections per reconcile loop.

Each reconciliation cycle was creating and tearing down multiple Redis client connections to the same pod IPs (e.g. in checkReplicaRole, isReplicaStable, replicaOf, replicaOfNoOne etc.). 

This adds unnecessary connection overhead and latency, especially in clusters with many replicas where several of these methods are called in sequence for the same pod.

### DIFF
--- a/internal/controller/dragonfly_controller.go
+++ b/internal/controller/dragonfly_controller.go
@@ -59,6 +59,7 @@ func (r *DragonflyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(fmt.Errorf("failed to get dragonfly instance: %w", err))
 	}
+	defer dfi.Close()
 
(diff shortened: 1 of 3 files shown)


miledxz:
> LGTM, but have you considered just closing connections immediately once operation is over, instead of keeping connection for entire reconcile iteration?

The cache map - `redisClients map[string]*redis.Client` 
lets those share one connection instead of each opening and closing their own. AFAIK reconcile cycles are short (seconds), so there's no real risk of holding connections.

Closing immediately is what the original implementation already does. The same pod often gets hit by multiple methods within one reconcile loop (e.g. first to check readiness, then to configure replication), so reusing the connection avoids repeated TCP handshakes to the same address. 

Also, everything gets cleaned up via defer when the reconcile returns.

Issue #465, opened 2026-02-25
Replica marked as ready while doing a full replicate from master


NegatioN:
Hi!
We're running a Dragonfly 1master/1replica setup, where there's no snapshotting at all, and strictly relying on live replication.
When a node gets drained by external actors (e.g. for a kubernetes update), with the current PodDisruptionBudget-setup, a single pod gets removed. Then, this this pod gets rescheduled, and comes up as `Ready` according to the Kubernetes healthchecks. Then it initiates a full replication process, while it still seems to be marked as `Ready`.
During this time, we can have the external actor continue draining nodes, and thereby killing the node the master is on, if replication takes long enough, because `maxUnavailable=1` Pods is satisfied.

This is similar to issue #397 , which was fixed recently.

I've been experimenting with some fixes:
1. Using `readinessGates` that mark pods while they are still replicating data (would slightly change the Pod Spec, and need increased privileges)
2. Reducing the `maxUnavailable` pods to 0 while a replica has not entered into a stable replication. (would require adjusting the PDB on the fly)
3. Not tried, but maybe one could implement it in the same way #397 was fixed, but I haven't looked into this

What do you see as the most desired solution to this problem? Is it something completely outside the scope of what I've written here, or would you accept a PR with any of the solutions?

NegatioN:
Since I've previously interfaced with @vyavdoshenko and @moredure on a very similar issue before ( https://github.com/dragonflydb/dragonfly-operator/pull/401 ). What do you think about a possible solution for this?

vyavdoshenko:
@NegatioN 
Great question — the changes should only be needed in the dragonfly-operator repo. The Dragonfly server itself already exposes all the necessary information via `INFO` commands:                                                                                                 
                  
  - `INFO replication` reports master_sync_in_progress, master_link_status, master_last_io_seconds_ago
  - `INFO persistence` reports loading, load_state

 The gap is in how this state is communicated to Kubernetes.

FYI:
@moredure @miledxz

ashotland:
Yep see isReplicaStable

custom readiness gate sounds right but I think we should feature flag it so users can upgrade operator version without auto enabling it (which will require extended permissions and trigger a rolling update of the pods) 

@moredure @miledxz WDYT? 

@NegatioN are you interested to work on this ?

NegatioN:
[...]

PR #467, opened 2026-02-26
feat: add replication-aware readiness gate to prevent data loss

miledxz:
Adding an opt-in custom Pod ReadinessGate (`dragonflydb.io/replication-ready`) that prevents Kubernetes from considering replica pods Ready while they are still performing full replication sync. This closes the gap where the PDB (`maxUnavailable=1`) allows draining the master node while a rescheduled replica hasn't finished syncing, causing data loss in setups without snapshotting.

Also added an update for `isReady()` to check `ContainersReady` instead of `PodReady` to prevent a possible "deadlock" where the operator cannot act on pods whose readiness gate hasn't satisfied yet. `ContainersReady` is a strict prerequisite of `PodReady` so they only differ when readiness gates are involved.

When the feature flag is enabled, the operator:
1. Injects a `dragonflydb.io/replication-ready` readiness gate into the
   StatefulSet pod template
2. Sets the condition to `True` immediately for master pods
3. Sets the condition to `False` for replica pods and polls
   `isReplicaStable()` every 5 seconds until replication is complete,
   then sets it to `True`
4. Clears the condition to `False` whenever a pod's role label is removed
   (failover path) to prevent stale state

Kubernetes and the PDB see pods with an unsatisfied gate as not-ready, which prevents voluntary disruption of other pods while replication is in progress.

## Test idea
- Deploy with 1 master + 1 replica (no snapshot), enable the feature flag, drain the replica node — confirm master node drain is blocked until replication completes
- Verify toggling `enableReplicationReadinessGate: true` triggers a rolling update and pods receive the readiness gate condition
- Verify with feature disabled: behavior identical to current release
- Verify single-replica (no replication) setup is unaffected
- Verify failover: kill master, confirm stale `replication-ready=True` is cleared before re-election


internal/controller/dragonfly_pod_lifecycle_controller.go
```diff
@@ -133,6 +135,25 @@ func (r *DfPodLifeCycleReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				stable = false
+			}
+			if err := dfi.patchReplicationReadyCondition(ctx, &pod, stable); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to patch replication ready condition: %w", err)
+			}
+			if !stable {
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+			}
+		}
+	}
```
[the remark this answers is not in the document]
miledxz: added e2e tests

miledxz:
Users who want to enable the replication-aware readiness gate need to set:

enableReplicationReadinessGate: true

in their Dragonfly CR. 
#465

internal/controller/dragonfly_instance.go
```diff
@@ -754,6 +772,50 @@ func (dfi *DragonflyInstance) isPodReady(ctx context.Context, pod *corev1.Pod) (
...
+func (dfi *DragonflyInstance) patchReplicationReadyCondition(ctx context.Context, pod *corev1.Pod, ready bool) error {
+	condType := corev1.PodConditionType(resources.ReplicationReadyConditionType)
+	desiredStatus := corev1.ConditionFalse
+	if ready {
+		desiredStatus = corev1.ConditionTrue
+	}
+
+	for _, c := range pod.Status.Conditions {
+		if c.Type == condType && c.Status == desiredStatus {
+			return nil
```
miledxz, line 795: if the condition already exists with the desired status, do nothing, 

avoids unnecessary API calls

internal/controller/dragonfly_instance.go
```diff
@@ -754,6 +772,50 @@ func (dfi *DragonflyInstance) isPodReady(ctx context.Context, pod *corev1.Pod) (
...
+			found = true
+			break
+		}
+	}
+	if !found {
+		pod.Status.Conditions = append(pod.Status.Conditions, corev1.PodCondition{
+			Type:               condType,
+			Status:             desiredStatus,
+			LastTransitionTime: now,
+		})
```
miledxz, line 818: either update the existing condition (if found) or append a new one

sets LastTransitionTime to now

internal/controller/dragonfly_instance.go
```diff
@@ -754,6 +772,50 @@ func (dfi *DragonflyInstance) isPodReady(ctx context.Context, pod *corev1.Pod) (
...
+	}
+
+	if err := dfi.client.Status().Patch(ctx, pod, patchFrom); err != nil {
```
miledxz, line 821: `patchFrom := client.MergeFrom(pod.DeepCopy()) `

wraps that snapshot and tells the Kubernetes client: "when I call Patch() later, compute a JSON merge patch by diffing the original (this snapshot) against the modified version

When `dfi.client.Status().Patch(ctx, pod, patchFrom)` runs, the client library diffs the modified pod against the stored snapshot and sends only the changed fields to the API server.

A merge patch only touches the fields we changed, so it's less likely to conflict and is cheaper on the API server.

internal/resources/resources.go
```diff
@@ -143,6 +143,12 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
 		},
 	}
 
+	if df.Spec.EnableReplicationReadinessGate {
+		statefulset.Spec.Template.Spec.ReadinessGates = []corev1.PodReadinessGate{
```
miledxz, line 148: adds the `dragonflydb.io/replication-ready` readiness gate to the pod template 

tells Kubernetes: "every pod created from this template has a custom readiness gate that must be satisfied before the pod counts as Ready"

internal/controller/dragonfly_pod_lifecycle_controller.go
```diff
@@ -133,6 +135,25 @@ func (r *DfPodLifeCycleReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		r.EventRecorder.Event(dfi.df, corev1.EventTypeNormal, "Replication", "Configured a new replica")
 	}
 
+	if dfi.df.Spec.EnableReplicationReadinessGate {
```
miledxz, line 139: every time a single pod event triggers reconciliation (pod created, updated, restarted, etc.), after the operator has finished configuring the pod's role, it evaluates the readiness condition

internal/controller/dragonfly_pod_lifecycle_controller.go
```diff
@@ -133,6 +135,25 @@ func (r *DfPodLifeCycleReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if isMaster(&pod) {
+			if err := dfi.patchReplicationReadyCondition(ctx, &pod, true); err != nil {
```
miledxz, line 141: always set to True , as master is always replication-ready

internal/controller/dragonfly_pod_lifecycle_controller.go
```diff
@@ -133,6 +135,25 @@ func (r *DfPodLifeCycleReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			stable, stableErr := dfi.isReplicaStable(ctx, &pod)
+			if stableErr != nil {
+				stable = false
+			}
+			if err := dfi.patchReplicationReadyCondition(ctx, &pod, stable); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to patch replication ready condition: %w", err)
+			}
+			if !stable {
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+			}
```
miledxz, line 154: check isReplicaStable(). 

If stable, set True. 

If not stable (sync in progress), set False and requeue after 5 seconds to check again. 

This adds polling loop that eventually flips the condition from False to True once replication finishes.

internal/controller/dragonfly_instance.go
```diff
@@ -811,6 +873,13 @@ func (dfi *DragonflyInstance) deleteRoleLabel(ctx context.Context, pod *corev1.P
 		return err
 	}
 
+	if dfi.df.Spec.EnableReplicationReadinessGate {
+		if err := dfi.patchReplicationReadyCondition(ctx, pod, false); err != nil {
```
miledxz, line 885: we set the condition to False because a pod without a role should not be considered replication-ready, it's transitioning

internal/controller/dragonfly_instance.go
miledxz, line 935: isReplicaStable() checks whether the newly updated replica has finished its full replication sync

internal/controller/dragonfly_instance.go
```diff
@@ -860,6 +929,12 @@ func (dfi *DragonflyInstance) verifyUpdatedReplicas(ctx context.Context, replica
 				return ctrl.Result{}, fmt.Errorf("failed to check if replica is stable: %w", err)
 			}
 
+			if dfi.df.Spec.EnableReplicationReadinessGate {
+				if patchErr := dfi.patchReplicationReadyCondition(ctx, &replica, ok); patchErr != nil {
```
miledxz, line 941: then patch the readiness condition to match result from isReplicaStable() check

NegatioN:
Is it possible to get a review on this @ashotland @moredure @tjovicic ? I'm understanding of the fact that you guys probably have a lot of things on your plate, but this PR would fix a huge issue for us, which leads to manual intervention every time a node need to get drained. :pray:

internal/controller/dragonfly_pod_lifecycle_controller.go
Abhra303: makes sense to include these as comments to explain the code.

miledxz:
> Is it possible to get a review on this @ashotland @moredure @tjovicic ? I'm understanding of the fact that you guys probably have a lot of things on your plate, but this PR would fix a huge issue for us, which leads to manual intervention every time a node need to get drained. 🙏

@NegatioN it's merged, please let us know did it help,

best !

NegatioN:
Wow, that's great to hear. Could we also do a release of the operator? :smile:

NegatioN:
@Abhra303 , is a release of the operator possible, so that this feature is accessible? :slightly_smiling_face:

NegatioN:
@miledxz , do you have any way to make the operator be released? :sweat_smile: We could obviously build the image locally and do it that way, but it would be nice to avoid.

miledxz:
> @miledxz , do you have any way to make the operator be released? 😅 We could obviously build the image locally and do it that way, but it would be nice to avoid.

sorry for later reply @NegatioN, we are in process of reviewing one more PR and hopefully releasing today or tomorrow, 

I will ping you when it happens

miledxz:
@NegatioN its released now, good luck !

PR #500, opened 2026-03-29
feat(operator): add custom probes in configmaps with override option

### LINKED ISSUE #499: Custom configmap-based health check probe scripts with override support
related: #401, https://github.com/dragonflydb/dragonfly/issues/5921

Currently CRD doesn't provide a way to customise pod readiness, liveness checks and startup probes - they are hardcoded in the operator binary. 

Following suggestions https://github.com/dragonflydb/dragonfly-operator/pull/401#issuecomment-3437514852 and https://github.com/dragonflydb/dragonfly/issues/5921#issuecomment-3409597930 opening an issue in `dragonfly-operator` repo with the PR to introduce custom pod readiness, liveness checks and startup probes

mvasilenko:
Problem:

Currently CRD doesn't provide a way to customise pod readiness, liveness checks and startup probes - they are hardcoded in the operator binary.

Following suggestions https://github.com/dragonflydb/dragonfly-operator/pull/401#issuecomment-3437514852 and https://github.com/dragonflydb/dragonfly/issues/5921#issuecomment-3409597930 opening PR to introduce custom pod readiness, liveness checks and startup probes.

Proposed solution:

  - 3 ConfigMaps per Dragonfly instance with the default probe scripts, mounted into pods at `/etc/dragonfly/probes/`
  - 3 new optional CR fields — `customLivenessProbeConfigMap`, `customReadinessProbeConfigMap`,
  `customStartupProbeConfigMap`
 - default scripts are embedded into the operator binary `//go:embed scripts/*.sh` and written into the generated
  ConfigMaps at reconcile time
  - RBAC ClusterRole updated to include configmaps CRUD verbs

Example:

custom `liveness` probe with connect timeout

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: dragonfly-sample-probes
  namespace: default
data:
  liveness-check.sh: |
    #!/bin/sh
    RESPONSE=$(timeout 4 redis-cli -h localhost -p ${HEALTHCHECK_PORT:-9999} PING 2>/dev/null)
    case "$RESPONSE" in
      PONG|*LOADING*) exit 0 ;;
      *)              exit 1 ;;
    esac
---
kind: Dragonfly
metadata:
  name: dragonfly-sample
spec:
  replicas: 1
  customLivenessProbeConfigMap:
    name: dragonfly-sample-probes
```


internal/resources/resources.go
```diff
@@ -398,14 +432,85 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
...
 		statefulset.Spec.Template.Spec.Volumes, df.Spec.AdditionalVolumes,
 		func(v corev1.Volume) string { return v.Name })
 
+	// ConfigMaps are appended before the StatefulSet so they exist when pods start
+	// and can mount the probe script volumes without getting stuck in Pending.
+	resources = append(resources,
+		generateProbeConfigMap(df, LivenessProbeConfigMapSuffix, LivenessScriptKey, defaultLivenessScript),
+		generateProbeConfigMap(df, ReadinessProbeConfigMapSuffix, ReadinessScriptKey, defaultReadinessScript),
+		generateProbeConfigMap(df, StartupProbeConfigMapSuffix, StartupScriptKey, defaultStartupScript),
+	)
```
miledxz: These ConfigMaps are the first resources without a .Spec field to enter the reconcile loop. The existing resourceSpecsEqual() in dragonfly_instance.go compares objects by FieldByName("Spec") https://github.com/dragonflydb/dragonfly-operator/blob/main/internal/controller/dragonfly_instance.go#L742 — when that field doesn't exist (as with ConfigMaps, which store content in .Data), it returns true ("equal") and skips the update. The patch path has the same issue.

Impact on that is next, 
if the operator ships updated probe scripts in a future release, existing ConfigMaps in the cluster will never be patched. Users would have to manually delete them.


This is a pre-existing gap in the reconciler, but this PR is the first to hit it. 

I suggest fixing resourceSpecsEqual and the patch block in reconcileResources to handle ConfigMap .Data before merging — otherwise probe script upgrades will silently fail on existing clusters.


Suggestion:

In `resourceSpecsEqual` in `dragonfly_instance`.go, add before the FieldByName("Spec") block:


```
if cmDesired, ok := desired.(*corev1.ConfigMap); ok {
   if cmExisting, ok := existing.(*corev1.ConfigMap); ok {
         return reflect.DeepEqual(cmDesired.Data, cmExisting.Data)
     }
 }

```

or something similar, 

WDYT @mvasilenko ?

internal/resources/resources.go
```diff
@@ -102,17 +125,19 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
 								},
 							},
 							Args: DefaultDragonflyArgs,
-							Env: append(df.Spec.Env, corev1.EnvVar{
-								Name:  "HEALTHCHECK_PORT",
-								Value: fmt.Sprintf("%d", DragonflyAdminPort),
```
miledxz, line 107: HEALTHCHECK_PORT was removed from the container env here, but the PR description's example for custom probes still references ${HEALTHCHECK_PORT:-6379}. Anyone following the example will have their custom probes silently hit port 6379 instead of the admin port 9999 — which defeats the purpose on TLS clusters.

we can re-add HEALTHCHECK_PORT=9999 as an env var so custom scripts can reference it portably,
it is more user-friendly — it gives custom script authors a stable abstraction rather than requiring them to know internal port numbers

internal/resources/resources.go
miledxz: also,

these three ConfigMaps are always generated even when the user has set customLivenessProbeConfigMap / customReadinessProbeConfigMap / customStartupProbeConfigMap. 

The custom override only changes the volume reference (lines 436-447), but the default ConfigMaps are still created, sit unused in the namespace, and get reconciled every loop.

My suggestion — skip generating the default when a custom override is provided:

```
 if df.Spec.CustomLivenessProbeConfigMap == nil {
     resources = append(resources, generateProbeConfigMap(df, LivenessProbeConfigMapSuffix, LivenessScriptKey, defaultLivenessScript))
 }
 if df.Spec.CustomReadinessProbeConfigMap == nil {
     resources = append(resources, generateProbeConfigMap(df, ReadinessProbeConfigMapSuffix, ReadinessScriptKey, defaultReadinessScript))
 }
 if df.Spec.CustomStartupProbeConfigMap == nil {
     resources = append(resources, generateProbeConfigMap(df, StartupProbeConfigMapSuffix, StartupScriptKey, defaultStartupScript))
 }
```

internal/resources/scripts/readiness-check.sh
```diff
@@ -0,0 +1,24 @@
...
+# respond to commands (including PING) regardless of loading state, so a PING
+# check would allow traffic to a pod that is still restoring a snapshot.
+#
+# Startup and liveness probes intentionally succeed during LOADING so the pod
+# is not restarted mid-restore. This probe is the sole gate on traffic.
+
+HOST="localhost"
+PORT=9999  # Dragonfly admin port — always plain-text, not user-configurable
+
+RESPONSE=$(redis-cli -h "$HOST" -p "$PORT" --no-auth-warning INFO persistence 2>/dev/null)
```
miledxz: Liveness and startup scripts pass ${DFLY_requirepass:+-a "$DFLY_requirepass"} to redis-cli, but the readiness script does not:

> ### readiness (no auth)
> RESPONSE=$(redis-cli -h "$HOST" -p "$PORT" --no-auth-warning INFO persistence 2>/dev/null)
>
> ### liveness (has auth)
> RESPONSE=$(redis-cli -h "$HOST" -p "$PORT" --no-auth-warning \
>   ${DFLY_requirepass:+-a "$DFLY_requirepass"} PING 2>/dev/null)
>

Even if the admin port currently doesn't require auth, the scripts should be consistent. If admin port auth behavior ever changes, only the readiness probe would break. I suggest adding the same ${DFLY_requirepass:+-a "$DFLY_requirepass"} to the readiness script as well.

internal/resources/resources.go
```diff
@@ -398,14 +432,85 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
...
+		},
+	)
+
 	statefulset.Spec.Template.Spec.Containers = mergeNamedSlices(
 		statefulset.Spec.Template.Spec.Containers, df.Spec.AdditionalContainers,
 		func(c corev1.Container) string { return c.Name })
 
+	// Note: AdditionalVolumes takes precedence — a volume named "liveness-probe",
+	// "readiness-probe", or "startup-probe" in spec.additionalVolumes will override
+	// the operator-generated probe volume for that slot.
```
miledxz: The comment here correctly notes that additionalVolumes can override the probe volumes by matching the volume name. Combined with the new customXxxProbeConfigMap API fields, users now have two different ways to override probe scripts, and nothing prevents them from using both at once (which could conflict).

maybe worth considering of adding a validation webhook that rejects the CR if both mechanisms target the same probe, or just documenting clearly in the CRD field descriptions which mechanism takes precedence.

miledxz (commented):
Overall looks good, left few comments @mvasilenko please take a look when you have time, 

sorry for later reply, was over busy this week :)

internal/resources/resources.go
```diff
@@ -102,17 +125,19 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
...
+							Env: df.Spec.Env,
+							// Probe semantics during dataset LOADING (large snapshot restore):
+						//   StartupProbe  — succeeds on any PING response (PONG or LOADING); prevents
+						//                   liveness from firing before the process is up.
+						//   LivenessProbe — succeeds on any PING response (PONG or LOADING); must NOT
+						//                   restart a pod that is mid-restore, as that aborts the load
+						//                   and creates a crash loop (see issues #426, #508).
+						//   ReadinessProbe — fails on LOADING; gates traffic until the dataset is fully
+						//                   loaded and Dragonfly can serve requests.
+						ReadinessProbe: &corev1.Probe{
```
miledxz: @mvasilenko worth running go fmt for this file

mvasilenko:
@miledxz addressed comments, appreciate re-review.

Thanks for your patience — this is my first contribution

Abhra303:
Hey, can you fix the conflicts please?

mvasilenko:
updated branch with new main, could you check please?

e2e/dragonfly_controller_test.go
```diff
@@ -136,10 +136,9 @@ var _ = Describe("Dragonfly Lifecycle tests", Ordered, FlakeAttempts(3), func()
 			err := k8sClient.Create(ctx, &df)
 			Expect(err).To(BeNil())
 
-			// Wait until Dragonfly object is marked initialized
-			waitForDragonflyPhase(ctx, k8sClient, name, namespace, controller.PhaseResourcesCreated, 2*time.Minute)
-			waitForStatefulSetReady(ctx, k8sClient, name, namespace, 2*time.Minute)
-
+			// Wait for master election, then for all replicas to be Ready
```
Abhra303: these e2e fix is already merged. Please update the PR with the latest main branch.

e2e/util.go
```diff
@@ -66,6 +66,28 @@ func parseTieredEntriesFromInfo(info string) (int64, error) {
 	return 0, fmt.Errorf("tiered_entries not found")
 }
 
+// waitForMasterPod polls until at least one pod with role=master exists. Use this
+// after waitForStatefulSetReady to guarantee the lifecycle controller has finished
+// master election before the test tries to connect.
+func waitForMasterPod(ctx context.Context, c client.Client, name, namespace string, maxDuration time.Duration) error {
```
Abhra303, line 72: these are merged. Please update the PR.

internal/resources/resources.go
```diff
@@ -33,6 +33,28 @@ var (
 	dflyUserGroup int64 = 999
 )
 
+func generateProbeConfigMap(df *resourcesv1.Dragonfly, suffix, key, script string) *corev1.ConfigMap {
```
Abhra303: pass configmap name instead of `suffix` and assign it to `Name` under ObjectMeta.

internal/resources/resources.go
```diff
@@ -103,16 +125,15 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
...
 								Value: fmt.Sprintf("%d", DragonflyAdminPort),
 							}),
 							ReadinessProbe: &corev1.Probe{
 								ProbeHandler: corev1.ProbeHandler{
 									Exec: &corev1.ExecAction{
-										Command: []string{
-											"/bin/sh",
-											"/usr/local/bin/healthcheck.sh",
-										},
+										Command: []string{"/bin/sh", ProbeMountPath + "/" + ReadinessScriptKey},
```
Abhra303: Instead of concatenating strings every time, maybe use a helper function that accepts probe file name.

internal/resources/resources.go
```diff
@@ -124,10 +145,7 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
 							LivenessProbe: &corev1.Probe{
 								ProbeHandler: corev1.ProbeHandler{
 									Exec: &corev1.ExecAction{
-										Command: []string{
-											"/bin/sh",
-											"/usr/local/bin/healthcheck.sh",
-										},
+										Command: []string{"/bin/sh", ProbeMountPath + "/" + LivenessScriptKey},
```
Abhra303: Instead of concatenating strings every time, maybe use a helper function that accepts probe file name.

internal/resources/resources.go
```diff
@@ -136,6 +154,18 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
 								SuccessThreshold:    1,
 								TimeoutSeconds:      5,
 							},
+							StartupProbe: &corev1.Probe{
+								ProbeHandler: corev1.ProbeHandler{
+									Exec: &corev1.ExecAction{
+										Command: []string{"/bin/sh", ProbeMountPath + "/" + StartupScriptKey},
```
Abhra303: Instead of concatenating strings every time, maybe use a helper function that accepts probe file name.

internal/resources/resources.go
```diff
@@ -398,14 +428,94 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
...
+			MountPath: ProbeMountPath + "/" + ReadinessScriptKey,
+			SubPath:   ReadinessScriptKey,
+		},
+		corev1.VolumeMount{
+			Name:      StartupProbeVolumeName,
+			MountPath: ProbeMountPath + "/" + StartupScriptKey,
+			SubPath:   StartupScriptKey,
+		},
+	)
+
```
Abhra303: could you please move all this to a helper function. The function is getting too big.

internal/resources/resources.go
```diff
@@ -398,14 +428,94 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
...
+	// Do not use both mechanisms for the same probe simultaneously.
 	statefulset.Spec.Template.Spec.Volumes = mergeNamedSlices(
 		statefulset.Spec.Template.Spec.Volumes, df.Spec.AdditionalVolumes,
 		func(v corev1.Volume) string { return v.Name })
 
+	// ConfigMaps are appended before the StatefulSet so they exist when pods start
+	// and can mount the probe script volumes without getting stuck in Pending.
+	// Skip generating a default when the user has pointed to their own ConfigMap.
+	if df.Spec.CustomLivenessProbeConfigMap == nil || df.Spec.CustomLivenessProbeConfigMap.Name == "" {
+		resources = append(resources, generateProbeConfigMap(df, LivenessProbeConfigMapSuffix, LivenessScriptKey, defaultLivenessScript))
```
Abhra303: pass the exact name i.e. fmt.Sprintf("%s-%s,....) here.

internal/resources/resources.go
```diff
@@ -398,14 +428,94 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
...
+	}
+	if df.Spec.CustomReadinessProbeConfigMap == nil || df.Spec.CustomReadinessProbeConfigMap.Name == "" {
+		resources = append(resources, generateProbeConfigMap(df, ReadinessProbeConfigMapSuffix, ReadinessScriptKey, defaultReadinessScript))
```
Abhra303: pass the exact name here instead of the suffix.

internal/resources/resources.go
```diff
@@ -398,14 +428,94 @@ func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage
...
+	}
+	if df.Spec.CustomStartupProbeConfigMap == nil || df.Spec.CustomStartupProbeConfigMap.Name == "" {
+		resources = append(resources, generateProbeConfigMap(df, StartupProbeConfigMapSuffix, StartupScriptKey, defaultStartupScript))
```
Abhra303: pass the exact name instead of the suffix.

internal/controller/dragonfly_controller.go
```diff
@@ -45,6 +45,7 @@ type DragonflyReconciler struct {
 //+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
```
Abhra303, line 48: You also need to add .Owns(&corev1.ConfigMap{},...) in `SetupWithManager` function in the dragonfly_controller.go. Else, reconciler won't be triggered when configmaps are deleted/updated.

Abhra303 (changes requested):
Hey, see some comments. Looks OK overall. But my main concern is when the operator is upgraded, all existing resources will be updated immediately as the operator reconciler will see the configmaps are missing and statefulset has different probe than desired. This is somewhat breaking the compatibility.

mvasilenko:
Thanks for the review, addressed comments.

As for the main concern, I've updated and tested the code - [see gist](https://gist.github.com/mvasilenko/96c016b73083fd1f2dc678a6272cb5b5), it should be no-op for existing users and only on CRD patch with opt-in for custom probes, pods would be recreated.

Another gap I found with current CM-based approach - on accidental probe's ConfigMap deletion, referenced in Dragonfly CR spec - the probes break and pods start failing. 

So we could add finalizers like `dragonflydb.io/probe-configmap-protection` to user-provided probe CMs or webhooks for CMs deletion prevention, wdyt?

internal/controller/dragonfly_instance.go
miledxz: configmap data updates are silently dropped meaning,

the configmap branch in resourceSpecsEqual (lines 748–753) correctly detects when data differs, so reconcile reaches the patch path, but the patch path only copies a Spec field via reflection:

```
desiredSpec := desiredV.FieldByName("Spec")
existingSpec := existingV.FieldByName("Spec")
if desiredSpec.IsValid() && existingSpec.IsValid() {
    existingSpec.Set(desiredSpec)
}
```

corev1.ConfigMap has no Spec, so existingSpec.IsValid() is false and existing.Data is never overwritten. client.Patch is then called with the unchanged data — the operator logs "patched resource" but nothing actually changes.

pls take a look @mvasilenko 

```suggestion
		if cmDesired, ok := desired.(*corev1.ConfigMap); ok {
			cmExisting := existing.(*corev1.ConfigMap)
			cmExisting.Data = cmDesired.Data
			cmExisting.BinaryData = cmDesired.BinaryData
		} else {
			desiredV := reflect.ValueOf(desired).Elem()
			existingV := reflect.ValueOf(existing).Elem()
			desiredSpec := desiredV.FieldByName("Spec")
			existingSpec := existingV.FieldByName("Spec")
			if desiredSpec.IsValid() && existingSpec.IsValid() {
				existingSpec.Set(desiredSpec)
			}
		}
```

also worth adding a regression test that mutates the desired CM and asserts the existing one is actually updated — current tests only cover the create path.

miledxz (commented):
one change required before LGTM, pls take a look @mvasilenko 

nice work !

miledxz (approved):

Abhra303 (approved):

PR #506, opened 2026-03-30
fix(controller): requeue reconciliation for replication readiness gate polling

### LINKED ISSUE #503: Bug: Readiness gate condition `dragonflydb.io/replication-ready` never transitions to `True` after replica sync completes
## Summary
 
When `enableReplicationReadinessGate: true` is enabled, the lifecycle controller sets the `dragonflydb.io/replication-ready` condition to `False` when configuring a replica pod, but never flips it to `True` after replication becomes stable. This is because the controller does not requeue the reconciliation — it checks `isReplicaStable()` once (while the replica is still mid-sync), sees `false`, and exits without scheduling a follow-up. Since no further pod events occur, the controller never reconciles again and the condition stays `False` permanently, blocking the rollout.
 
## Environment
 
- **Operator version:** v1.5.0
- **Dragonfly version:** v1.29.0
- **Kubernetes:** kind (tested locally) / GKE (production)
- **Replicas:** 3
 
## Introduced in
 
PR #467 — `feat: add replication-aware readiness gate to prevent data loss` (v1.5.0).
 
## Description
 
PR #467 describes the intended behavior:
 
> Sets the condition to `False` for replica pods and polls `isReplicaStable()` every 5 seconds until replication is complete, then sets it to `True`
 
However, the polling never happens. The flow is:
 
1. Lifecycle controller configures a replica: clears `snapshot_cron`, issues `SLAVE OF`, applies `role=replica` label, sets `replication-ready: False`.
2. The label change triggers a pod event, which causes one reconciliation of the lifecycle controller.
3. The controller sees the pod is a replica, calls `isReplicaStable()` — returns `false` because the replica is mid-sync.
4. The controller returns `ctrl.Result{}` (no requeue) and exits.
5. No further pod events occur (pod is Running, labeled, nothing changes from Kubernetes' perspective).
6. The controller never reconciles this pod again. The condition stays `False` forever.
 
Meanwhile, the rollout controller logs `"waiting for all replicas to be configured"` with `expected: 2, current: 1` indefinitely — it sees the replica pod as not Ready because the readiness gate is unsatisfied.
 
## Evidence
 
Replication is confirmed stable:
 
```
kubectl exec dragonfly-snapshot-0 -- redis-cli -p 9999 INFO replication | grep -E "role|master_link_status|master_sync_in_progress"
role:slave
master_link_status:up
master_sync_in_progress:0
```
 
But the readiness gate condition remains `False`:
 
```json
{
  "lastProbeTime": null,
  "lastTransitionTime": "2026-03-30T07:43:45Z",
  "status": "False",
  "type": "dragonflydb.io/replication-ready"
}
```
 
## Root cause
 
In `internal/controller/dragonfly_pod_lifecycle_controller.go`, the reconcile path for existing replica pods checks `isReplicaStable()` but does not return a `ctrl.Result{RequeueAfter: 5 * time.Second}` when the result is `false`. Without the requeue, the controller relies on external pod events to trigger re-reconciliation, but no such events occur after the role label is applied.
 
## Suggested fix
 
When `enableReplicationReadinessGate` is enabled and the pod is a replica with `replication-ready: False`, the controller should:
 
1. Call `isReplicaStable()`
2. If stable: patch condition to `True`, return `ctrl.Result{}`
3. If not stable: return `ctrl.Result{RequeueAfter: 5 * time.Second}` to keep polling
 
This matches the behavior described in the PR #467 description.
 
## Workaround
 
Manually patch the condition:
 
```bash
kubectl patch pod <pod-name> --subresource=status --type=merge -p \
  '{"status":{"conditions":[{"type":"dragonflydb.io/replication-ready","status":"True","lastTransitionTime":"'$(date -u +%Y-%m-%dT%H:%M:%SZ)'"}]}}'
```
 
## Related
 
- PR #467 — https://github.com/dragonflydb/dragonfly-operator/pull/467
- Issue #497 — https://github.com/dragonflydb/dragonfly-operator/issues/497 (readiness gate deadlock during master election)
- Issue #498 — https://github.com/dragonflydb/dragonfly-operator/issues/498 (snapshot_cron LOADING race)

davidcharbonnier:
Closes #503

### DIFF
--- a/internal/controller/dragonfly_pod_lifecycle_controller.go
+++ b/internal/controller/dragonfly_pod_lifecycle_controller.go
@@ -134,41 +134,37 @@ func (r *DfPodLifeCycleReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 	}
 
-	if !podReady {
-		return ctrl.Result{}, nil
-	}
+	if podReady {
+		if roleExists(&pod) {
+			if dfi.getStatus().Phase == PhaseReady || dfi.getStatus().Phase == PhaseReadyOld {
+				// is something wrong? check if all replicas have a matching role and revamp accordingly
+				log.Info("non-deletion event for a pod with an existing role. checking if something is wrong", "pod", pod.Name, "role", pod.Labels[resources.RoleLabelKey])
 
-	if roleExists(&pod) {
-		if dfi.getStatus().Phase != PhaseReady && dfi.getStatus().Phase != PhaseReadyOld {
-			return ctrl.Result{}, nil
-		}
+				if err = dfi.checkAndConfigureReplicas(ctx, master.Status.PodIP); err != nil {
+					// Check for transient replication errors
+					if isReplicationCancelledError(err) {
+						log.Info("replication cancelled (transient), will retry", "error", err)
+						return ctrl.Result{RequeueAfter: 3 * time.Second}, nil
+					}
+					return ctrl.Result{}, fmt.Errorf("failed to check and configure replicas: %w", err)
+				}
 
-		// is something wrong? check if all replicas have a matching role and revamp accordingly
-		log.Info("non-deletion event for a pod with an existing role. checking if something is wrong", "pod", pod.Name, "role", pod.Labels[resources.RoleLabelKey])
-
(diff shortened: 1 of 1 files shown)


davidcharbonnier:
I've opened 3 very close issues and 2 MRs that overlaps, will close this one and reopen a single MR to fix all 3 issues, sorry for this

PR #507, opened 2026-03-30
fix: ensure only one replica is deleted at a time during rolling update

### LINKED ISSUE #504: Bug: Rolling update deletes all replica pods simultaneously instead of one at a time
## Summary
 
During a rolling update, the operator's rollout controller deletes all replica pods with stale specs in a single reconcile pass instead of one at a time. For a 3-replica cluster (1 master + 2 replicas), both replicas are deleted within ~1 second of each other, leaving the master with zero connected replicas during the entire rollout window. This defeats the purpose of a rolling update and makes the PDB ineffective.
 
## Environment
 
- **Operator version:** v1.5.0
- **Dragonfly version:** v1.29.0
- **Kubernetes:** kind (tested locally) / GKE (production)
- **Replicas:** 3
 
## Description
 
When a pod spec change is detected (e.g., toggling `enableReplicationReadinessGate`, changing resource limits, bumping image version), the rollout controller:
 
1. Identifies all pods with stale specs
2. Deletes **all** replica pods in a single pass
3. Waits for the STS to recreate them and the lifecycle controller to configure them
4. Then proceeds to roll the master last
 
The expected behavior for a rolling update is to delete one replica at a time, wait for it to be fully configured and synced with the master, then proceed to the next. This ensures at least one replica is always available for failover.
 
## Operator logs
 
```
07:42:57Z  INFO  pod spec has changed, performing a rollout
07:42:57Z  INFO  deleting replica    pod: dragonfly-snapshot-0
07:42:57Z  INFO  deleting replica    pod: dragonfly-snapshot-0   (retry — pod still terminating)
07:42:58Z  INFO  deleting replica    pod: dragonfly-snapshot-2   (second replica deleted <1s later)
```
 
Both replicas are deleted before either has been recreated, configured, or synced.
 
## Impact
 
- **Data loss risk:** If the master pod dies while both replicas are being recreated, there is no synced replica to fail over to. With `snapshot.cron` configured, the last snapshot may be minutes old.
- **PDB bypass:** The PDB (`maxUnavailable: 1`) is designed to prevent more than one pod from being unavailable at a time, but the operator bypasses the PDB by directly deleting pods via the API rather than through eviction.
- **Readiness gate interaction:** With `enableReplicationReadinessGate: true`, the rollout controller then waits for both replicas to come back and satisfy their readiness gates. Since both are syncing simultaneously from the master, this doubles the replication load on the master and extends the window of vulnerability.
 
## Expected behavior
 
The rollout controller should:
 
1. Delete one replica pod
2. Wait for the STS to recreate it
3. Wait for the lifecycle controller to configure it as a replica (apply role label, set up replication)
4. Optionally, if `enableReplicationReadinessGate` is enabled, wait for the readiness gate to be `True` (replication stable)
5. Then proceed to delete the next replica pod
6. After all replicas are rolled, handle the master (failover + roll)
 
This ensures at least one synced replica is always available during the rollout.
 
## Suggested fix
 
In the rollout controller (`internal/controller/dragonfly_controller.go` or equivalent), the replica deletion loop should:
 
1. Delete only one replica pod per reconcile iteration
2. Return `ctrl.Result{RequeueAfter: ...}` to check back after the pod is recreated and configured
3. Only proceed to the next replica when the previous one has `role=replica` label and (if readiness gate is enabled) `replication-ready: True`
 
## Related
 
- Issue #497 — https://github.com/dragonflydb/dragonfly-operator/issues/497 (readiness gate deadlock)
- Issue #498 — https://github.com/dragonflydb/dragonfly-operator/issues/498 (snapshot_cron LOADING race)
- PR #422 — `fix: rolling update process` (v1.4.0) — attempted to fix rolling update logic but did not address simultaneous replica deletion

davidcharbonnier:
Fixes #504

### DIFF
--- a/internal/controller/dragonfly_instance.go
+++ b/internal/controller/dragonfly_instance.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -929,6 +930,11 @@ func (dfi *DragonflyInstance) allPodsHealthyAndHaveRole(ctx context.Context, upd
 		return ctrl.Result{}, fmt.Errorf("failed to get dragonfly pods: %w", err)
 	}
 
+	// Sort pods by name to ensure deterministic deletion order across concurrent reconciles.
+	sort.Slice(pods.Items, func(i, j int) bool {
+		return pods.Items[i].Name < pods.Items[j].Name
+	})
+
 	for _, pod := range pods.Items {
 		// Ignore pods that are already terminating
 		if isTerminating(&pod) {
@@ -997,6 +1003,12 @@ func (dfi *DragonflyInstance) updateReplicas(ctx context.Context, replicas *core
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to get master before deleting replica: %w", err)
 	}
+
+	// Sort replicas by name to ensure deterministic deletion order across concurrent reconciles.
+	// This prevents a race condition where multiple reconciles could delete different pods simultaneously.
+	sort.Slice(replicas.Items, func(i, j int) bool {
+		return replicas.Items[i].Name < replicas.Items[j].Name
+	})
 
 	// Check if any replica is already terminating.
 	// We want to delete only 1 replica at a time to ensure we have enough replicas for master failover.


internal/controller/dragonfly_instance.go
```diff
@@ -929,6 +930,11 @@ func (dfi *DragonflyInstance) allPodsHealthyAndHaveRole(ctx context.Context, upd
 		return ctrl.Result{}, fmt.Errorf("failed to get dragonfly pods: %w", err)
 	}
 
+	// Sort pods by name to ensure deterministic deletion order across concurrent reconciles.
```
ashotland, line 933: @davidcharbonnier why do you have concurrent reconcilers ?

are you setting MaxConcurrentReconciles > 1 ?

davidcharbonnier: No, I'm not setting `MaxConcurrentReconciles > 1`. The comment wording is misleading. The sort is just defensive to ensure deterministic order regardless of how `client.List` returns pods.

The actual fix here is the terminating-pod check below it: if any replica is already terminating, we return early with a requeue instead of continuing the loop and deleting the next one. Without this, a single reconcile pass iterates over all outdated replicas and deletes them all before any replacement has a chance to come back.

PR #514, opened 2026-04-08
fix(security): harden default NetworkPolicy for client and admin ports

miledxz:
Fixes the issue #509  where the default NetworkPolicy was effectively granting
unrestricted access to port 6379 and allowing any pod with the
`control-plane: controller-manager` label to reach the admin port from
any namespace.

- **Restrict port 6379 (client) and memcached port** to same-namespace pods only,
  instead of being open to the entire cluster
- **Pin admin port (9999) operator access** to the operator's actual namespace using
  the `kubernetes.io/metadata.name` well-known label, preventing cross-namespace
  access via spoofed pod labels
- **Secure fallback**: when the operator namespace cannot be determined, the admin
  port defaults to same-namespace-only access
- **Add `POD_NAMESPACE` env var** to both Helm chart and kustomize deployment
  manifests via the Downward API for reliable namespace detection


cmd/main.go
```diff
@@ -227,3 +236,21 @@ func addNamespacesToOpts(namespaces string, ops *ctrl.Options) error {
...
+
+func getOperatorNamespace() string {
+	return resolveOperatorNamespace(saNamespaceFile)
+}
+
+func resolveOperatorNamespace(saFile string) string {
+	if ns, ok := os.LookupEnv("POD_NAMESPACE"); ok && ns != "" {
+		return ns
+	}
+	if data, err := os.ReadFile(saFile); err == nil {
```
Abhra303, line 250: did you test this? (if POD_NAMESPACE is not available)

Abhra303 (approved):
LGTM

komapa:
👋🏼 anything still blocking this PR? It would be great to close this security issue with the operator!

miledxz:
> 👋🏼 anything still blocking this PR? It would be great to close this security issue with the operator!

@komapa will resolve comments and merge PR, thank you for heads up

cmd/main.go
miledxz: Yes — the "POD_NAMESPACE not available" path is covered. resolveOperatorNamespace() was extracted specifically so it's unit-testable, and cmd/main_test.go has table-driven cases for it. 

commit: https://github.com/dragonflydb/dragonfly-operator/pull/514/changes/0105df4447eea8d1ae408070a7b3d8496d055f4c

miledxz:
@copilot resolve the merge conflicts in this pull request

PR #516, opened 2026-04-08
fix(e2e): wait for PhaseReady before accessing master pod

mvasilenko:
Single-replica tests waited on PhaseResourcesCreated, which is set before the lifecycle controller runs configureReplication and labels the master pod, racing checkAndK8sPortForwardRedis with master election and intermittently failing with "no master pod found".


Abhra303 (approved):
LGTM, Thanks for the fix! 🚀

PR #517, opened 2026-04-11
fix(api): add podMetadata field and protect operator-managed labels

Abhra303:
Introduce spec.podMetadata (labels + annotations) as the replacement for the ambiguously named spec.labels and spec.annotations, which only ever applied to pod templates. spec.labels and spec.annotations are now deprecated in favour of the new field; both are still read for backward compatibility with podMetadata taking precedence when both are set.

Rename the OwnedObjectsMetadata type to MetadataSpec so it reads naturally as the shared type for both spec.ownedObjectsMetadata and spec.podMetadata. JSON field names are unchanged so existing manifests are unaffected.

Add label protection across all four user-facing label fields (spec.labels, spec.podMetadata.labels, spec.ownedObjectsMetadata.labels, spec.serviceSpec.labels): any key that collides with an operator-managed label (app.kubernetes.io/* or app) is rejected with a clear error pointing at the offending field and key. This fixes a crash where spec.labels: {app.kubernetes.io/name: custom} caused StatefulSet creation to fail with "selector does not match template labels".

Also fix spec.serviceSpec.{labels,annotations} to merge into the operator-generated labels rather than replacing them entirely.


internal/resources/resources.go
```diff
@@ -33,9 +34,50 @@ var (
...
+			return fmt.Errorf("%s: cannot override operator-managed label %q", field, key)
+		}
+	}
+	return nil
+}
+
 // GenerateDragonflyResources returns the resources required for a Dragonfly
 // Instance
 func GenerateDragonflyResources(df *resourcesv1.Dragonfly, defaultDragonflyImage string) ([]client.Object, error) {
+	if err := checkLabels("spec.labels", df.Spec.Labels); err != nil {
```
miledxz, line 62: I think `checkLabels` is silently skipped for nil maps but still called on nil for spec.labels 

meaning, spec.labels isn't nil-guarded before calling checkLabels

suggestion:
```suggestion
	if df.Spec.Labels != nil {
    if err := checkLabels("spec.labels", df.Spec.Labels); err != nil {
        return nil, err
    }
}
```

miledxz (commented):
nit: 

spec.podMetadata.annotations and spec.ownedObjectsMetadata.annotations are not validation protected, but maybe not needed ? WDYT @Abhra303 

question:

Currently AFIU the label protection runs in `GenerateDragonflyResources`, which is called during reconciliation. Meaning the invalid CR is accepted by the API server and the error only surfaces in the controller logs. 

Adding a validating webhook would give the user immediate feedback on kubectl apply. 

This is rather a design choice rather than a bug, WDYT ? no need to update PR, just out of curiosity. 

LGTM 🚀

internal/resources/resources.go
Abhra303: for nil labels, its a no-op function. The check is unnecessary IMO.

Abhra303:
> nit:
> 
> spec.podMetadata.annotations and spec.ownedObjectsMetadata.annotations are not validation protected, but maybe not needed ? WDYT @Abhra303

They are just strings. I don't think they need any validation.

> question:
> 
> Currently AFIU the label protection runs in `GenerateDragonflyResources`, which is called during reconciliation. Meaning the invalid CR is accepted by the API server and the error only surfaces in the controller logs.
> 
> Adding a validating webhook would give the user immediate feedback on kubectl apply.

Hmm, good point. Worth implementing it in next version release.

api/v1alpha1/dragonfly_types.go
```diff
@@ -58,16 +58,21 @@ type DragonflySpec struct {
...
+	// Deprecated: use spec.podMetadata.labels instead.
 	// +optional
 	// +kubebuilder:validation:Optional
 	Labels map[string]string `json:"labels,omitempty"`
 
+	// (Optional) Labels and annotations to add to the Dragonfly pods.
+	// +optional
+	// +kubebuilder:validation:Optional
+	PodMetadata *MetadataSpec `json:"podMetadata,omitempty"`
+
```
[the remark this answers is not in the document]
Abhra303: I will modify the manifest in next PR.

miledxz (approved):
lgtm

PR #518, opened 2026-04-11
fix(make): use server side apply to avoid hitting annotation max limit

Abhra303:
Client side kubectl apply tries to fit the last configured state of the dragonfly crd in an annotation. This makes the apply fail as annotation hits max limit.


miledxz (approved):
LGTM

PR #519, opened 2026-04-11
fix(controller): ignore immutable field modifications for statefulset

### LINKED ISSUE #472: bug: resize pvc sts template is not possible
Version: 1.4.0

Updates to persistentVolumeClaimSpec will result in a failed reconciliation since the change is updated directly in the statefulset pvc template which is not allowed.
There are probably also other changes which lead to the same problem with the current controller design:
```
2026-03-06T09:24:52Z	ERROR	Reconciler error	{"controller": "Dragonfly", "controllerGroup": "dragonflydb.io", "controllerKind": "Dragonfly", "Dragonfly": {"name":"dragonfly-provider","namespace":"devbox-sre-3"}, "namespace": "devbox-sre-3", "name": "dragonfly-provider", "reconcileID": "8471df95-a718-4eea-adf1-59cfcec2f2be", "error": "failed to reconcile dragonfly resources: failed to update resource: StatefulSet.apps \"dragonfly-provider\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:347
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:294
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:255`
```

The easiest solution is likely that the controller does an orphan deletion of the sts and recreate it with the new specs. Also the changes to the pvc template would need to be added to pvc's created by the statefulset controller directly.

Alternative would be not to use statefulset at all.

Abhra303:
Statefulset has some fields that are immutable e.g. volumeClaims. Any attempt to update these fields would fail causing the controller to be stuck in update state. Ignore these fields when update to the desired state. The controller will throw warnings if user tries to update immutable fields.

Note, this PR doesn't create a new statefulset to update the immutable fields. I think it would best to leave it for users to create a new Dragonfly resource with updated pvcs etc. i.e. making it behave identically with statefulset. Allowing updating these fields would need sts deletion, handling orphan pods etc. Manual handling of such cases makes more sense to me than by the dragonfly operator.

### DIFF
--- a/internal/controller/dragonfly_instance.go
+++ b/internal/controller/dragonfly_instance.go
@@ -647,6 +647,25 @@ func (dfi *DragonflyInstance) reconcileResources(ctx context.Context) error {
 		if err := controllerutil.SetControllerReference(dfi.df, desired, dfi.scheme); err != nil {
 			return fmt.Errorf("failed to set controller reference: %w", err)
 		}
+		// Special handling for StatefulSets to preserve immutable fields.
+		// VolumeClaimTemplates, Selector, ServiceName, and PodManagementPolicy
+		// cannot be changed after creation. Copy them from existing so they
+		// never appear as a diff. If the user changes e.g. persistentVolumeClaimSpec,
+		// the change is silently dropped here — a delete+recreate is required.
+		if stsDesired, ok := desired.(*appsv1.StatefulSet); ok {
+			if stsExisting, ok := existing.(*appsv1.StatefulSet); ok {
+				if !reflect.DeepEqual(stsDesired.Spec.VolumeClaimTemplates, stsExisting.Spec.VolumeClaimTemplates) {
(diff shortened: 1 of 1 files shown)


miledxz (approved):
lgtm, nice

PR #529, opened 2026-05-05
fix(controller): pod role label updates

bwnuts:
Make pod role label updates resilient to API conflicts and dataset-load delays

Three independent but related issues left dragonfly pods stuck without a
role=master/replica label:

1. A lost label Patch (e.g. due to a transient k8s API outage at the
   moment the pod was created) was never recovered. The pod kept
   replicating on the Dragonfly side but had no role label on the k8s
   side, and no later reconcile would re-apply it.

2. The lifecycle reconciler returned early on !podReady — which is true
   for the entire duration of a snapshot load or full sync. Since
   SLAVE OF is what *starts* the load, gating it on "load finished"
   was a self-deadlock: unlabeled pods loading from a snapshot stayed
   unlabeled forever.

3. Label/annotation updates were sent via plain client.Patch against
   the in-memory pod captured at reconcile entry. Concurrent kubelet
   status writes (probes, IPs, conditions) regularly bumped the
   resourceVersion before our patch landed, producing:

       Operation cannot be fulfilled on pods "...": the object has
       been modified; please apply your changes to the latest version
       and try again

Changes:

  * util.go: add isReachable() — Used to gate replication-topology mutations that only need the admin socket.

  * dragonfly_instance.go: add patchPodMetadata() that re-GETs the pod
    inside retry.RetryOnConflict, applies a MergeFrom patch and treats
    IsNotFound as success. All label/annotation writers (replicaOf,
    replicaOfNoOne, replTakeover, deleteRoleLabel) now route through it.

  * dragonfly_instance.go: add reconcileReplicaLabel() — if Redis
    already shows the correct master via INFO replication, only patch
    the missing label; do not re-issue SLAVE OF.

  * dragonfly_instance.go: configureReplication and checkAndConfigureReplicas 
    now iterate pods with isReachable instead of isPodReady.

  * dragonfly_pod_lifecycle_controller.go: move the "no role label →
    reconcile as replica" branch above the !podReady early return,
    gated on iReachable, calling reconcileReplicaLabel. Transient
    "replication cancelled" errors now requeue rather than surfacing
    as reconcile errors.

Verified locally on a Kind cluster across four scenarios: stripped role
label restoration, snapshot-load while unlabeled, rolling update with
concurrent annotation churn, and master-failover.

### DIFF
--- a/internal/controller/dragonfly_instance.go
+++ b/internal/controller/dragonfly_instance.go
@@ -92,7 +92,7 @@ func (dfi *DragonflyInstance) Close() {
 
 // patchPodMetadata patches a pod's labels/annotations with retry-on-conflict.
 func (dfi *DragonflyInstance) patchPodMetadata(ctx context.Context, podKey types.NamespacedName, mutate func(*corev1.Pod)) error {
-	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		var pod corev1.Pod
 		if err := dfi.client.Get(ctx, podKey, &pod); err != nil {
 			if apierrors.IsNotFound(err) {
@@ -102,7 +102,7 @@ func (dfi *DragonflyInstance) patchPodMetadata(ctx context.Context, podKey types
 			return err
 		}
 
-		patch := client.MergeFrom(pod.DeepCopy())
+		patch := client.MergeFromWithOptions(pod.DeepCopy(), client.MergeFromWithOptimisticLock{})
 		if pod.Labels == nil {
 			pod.Labels = map[string]string{}
 		}
@@ -215,6 +215,8 @@ func (dfi *DragonflyInstance) reconcileReplicaLabel(ctx context.Context, pod *co
 			p.Annotations[resources.MasterIpAnnotationKey] = sanitized
 			if ip := net.ParseIP(sanitized); ip != nil && ip.To4() != nil {
 				p.Labels[resources.MasterIpLabelKey] = sanitized
+        } else {
+          delete(p.Labels, resources.MasterIpLabelKey)
 			}
 		})
 	}
@@ -337,7 +339,7 @@ func (dfi *DragonflyInstance) checkAndConfigureReplicas(ctx context.Context, mas
 			}
 		}
 
-		// Label unlabeled pods even while loading a snapshot.
+		// Configure (SLAVE OF) and label unlabeled pods even while loading a snapshot.
 		if !roleExists(&pod) && isReachable(&pod) {
 			if err := dfi.configureReplica(ctx, &pod, masterIp); err != nil {
 				return err
--- a/internal/controller/dragonfly_instance.go
+++ b/internal/controller/dragonfly_instance.go
@@ -570,7 +570,6 @@ func (dfi *DragonflyInstance) replicaOf(ctx context.Context, pod *corev1.Pod, ma
 	return nil
 }
 
-
 // replicaOfNoOne configures the pod as a master along while updating other pods to be replicas
 func (dfi *DragonflyInstance) replicaOfNoOne(ctx context.Context, pod *corev1.Pod) error {
 	redisClient := dfi.getRedisClient(pod.Status.PodIP)
--- a/internal/controller/dragonfly_instance.go
+++ b/internal/controller/dragonfly_instance.go
@@ -91,11 +91,6 @@ func (dfi *DragonflyInstance) Close() {
 }
 
 // patchPodMetadata patches a pod's labels/annotations with retry-on-conflict.
-// Re-GETs the pod inside the retry loop to avoid stale-resourceVersion 409s
-// from concurrent kubelet status writes. NotFound is treated as success so
-// the caller stays idempotent when the pod was deleted between the Redis
-// command and the label update. The mutate callback runs on a fresh pod and
-// must not call the k8s client.
 func (dfi *DragonflyInstance) patchPodMetadata(ctx context.Context, podKey types.NamespacedName, mutate func(*corev1.Pod)) error {
 	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		var pod corev1.Pod
@@ -164,15 +159,14 @@ func (dfi *DragonflyInstance) configureReplication(ctx context.Context, master *
 		}
 	}
 
-	// Use isPodReachable rather than isPodReady: SLAVE OF only needs the
-	// admin socket and is itself what starts the dataset load. The
-	// readiness gate below still waits for stable sync before admitting traffic.
+	// SLAVE OF only needs the admin socket and is what starts the dataset load,
+	// so gate on isReachable rather than isPodReady.
 	for _, pod := range pods.Items {
 		if pod.Name == master.Name {
 			continue
 		}
 
-		if !isPodReachable(&pod) {
+		if !isReachable(&pod) {
 			dfi.log.Info("skipping replica configuration; pod not reachable yet", "pod", pod.Name)
 			continue
 		}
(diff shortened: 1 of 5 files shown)


internal/controller/dragonfly_instance.go
```diff
@@ -88,6 +90,37 @@ func (dfi *DragonflyInstance) Close() {
...
+		var pod corev1.Pod
+		if err := dfi.client.Get(ctx, podKey, &pod); err != nil {
+			if apierrors.IsNotFound(err) {
+				dfi.log.Info("pod disappeared before metadata update; skipping", "pod", podKey.String())
+				return nil
+			}
+			return err
+		}
+
+		patch := client.MergeFrom(pod.DeepCopy())
```
miledxz: a plain client.MergeFrom produces a JSON merge patch without resourceVersion, so the API server can never return 409 Conflict for it, which means the retry.RetryOnConflict wrapper on line 95 is effectively dead code and will never retry. 

the conflict resilience mentioned in the commit message isn't actually delivered by the retry; only the fresh Get helps

Now this also leaves a last-write-wins race between concurrent writers: reconciler A reads the pod, B patches role=master, then A patches role=replica and silently clobbers B. To get real read-modify-write semantics, use the optimistic-lock variant so 409s can occur and the retry loop becomes meaningful:

`		patch := client.MergeFromWithOptions(pod.DeepCopy(), client.MergeFromWithOptimisticLock{})
`

code suggestion:

- replace line 95 with 	`return retry.RetryOnConflict(retry.DefaultRetry, func() error {`
- replace line 105 with 	 `patch := client.MergeFromWithOptions(pod.DeepCopy(), client.MergeFromWithOptimisticLock{})`


WDYT @bwnuts ?

internal/controller/dragonfly_instance.go
```diff
@@ -170,6 +202,28 @@ func (dfi *DragonflyInstance) configureReplica(ctx context.Context, pod *corev1.
...
+	if err == nil && correct {
+		dfi.log.Info("Redis already replicating from correct master; only patching missing label", "pod", pod.Name, "master", masterIp)
+		sanitized := sanitizeIp(masterIp)
+		podKey := types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
+		return dfi.patchPodMetadata(ctx, podKey, func(p *corev1.Pod) {
+			p.Labels[resources.RoleLabelKey] = resources.Replica
+			p.Annotations[resources.MasterIpAnnotationKey] = sanitized
+			if ip := net.ParseIP(sanitized); ip != nil && ip.To4() != nil {
+				p.Labels[resources.MasterIpLabelKey] = sanitized
+			}
```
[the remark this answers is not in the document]
miledxz: @bwnuts worth checking this one

miledxz (changes requested):
left comments, please take a look

bwnuts:
@miledxz, thank you for the review!
Addressed the review comments in this update:
- patchPodMetadata now uses `MergeFromWithOptimisticLock` + `RetryOnConflict(retry.DefaultRetry)`.
- In `reconcileReplicaLabel`, added cleanup of stale `MasterIpLabelKey` for non-IPv4 paths (matching `replicaOf` behavior).
- Updated the unlabeled-pod comment in `checkAndConfigureReplicas` to reflect actual behavior.

bwnuts:
@miledxz, could you please take a look, thanks

internal/controller/dragonfly_instance.go
```diff
@@ -170,6 +202,30 @@ func (dfi *DragonflyInstance) configureReplica(ctx context.Context, pod *corev1.
 	return nil
 }
 
+// reconcileReplicaLabel idempotently re-applies the role=replica label.
+// If Redis is already replicating from the right master, only the label is patched.
```
miledxz: please update Redis to Dragonfly in comment :)

internal/controller/dragonfly_instance.go
```diff
@@ -170,6 +202,30 @@ func (dfi *DragonflyInstance) configureReplica(ctx context.Context, pod *corev1.
...
+			p.Annotations[resources.MasterIpAnnotationKey] = sanitized
+			if ip := net.ParseIP(sanitized); ip != nil && ip.To4() != nil {
+				p.Labels[resources.MasterIpLabelKey] = sanitized
+			} else {
+				delete(p.Labels, resources.MasterIpLabelKey)
+			}
+		})
+	}
+	if err != nil {
+		dfi.log.Info("could not inspect Redis replication state; falling back to full configureReplica", "pod", pod.Name, "error", err.Error())
```
miledxz: please update Redis to Dragonfly here :)

miledxz (changes requested):
@bwnuts one more test and we are done,

go test ./internal/controller passes, so the controller package still compiles and its current unit tests are green, 

the missing piece is behavioral coverage for the actual bug this PR is fixing, the changes to reconciliation so a pod can recover its role=replica label as soon as the Dragonfly admin socket/container is reachable, even if the dataset is still loading and the pod is not fully “ready” by the operator’s stricter dataset-loaded check

existing dataset-loading e2e test does not prove that behavior, it only verifies the pod eventually has a role label after loading finishes. That would pass even with the old broken behavior, where loading pods stayed unlabeled until dataset loading completed. So the test does not guard against a regression of this fix.

A useful test would assert this intermediate state:

- delete/restart a pod with persisted data
- wait until it is Running/container-ready/admin reachable but dataset loading is still in progress
- assert the pod already has role=replica
- assert Dragonfly/PodReady may still be blocked until loading completes

bwnuts:
@miledxz, thank you for the review!
The following changes were made:
1. New e2e regression test
   - Added `e2e/dragonfly_role_label_loading_test.go`.
   - Test validates the intermediate state:
     - restart pod with persisted data,
     - wait until pod is `Running` and dragonfly container/admin socket is reachable,
     - verify `role=replica` is present **while loading is still in progress**,
     - verify `PodReady` can still be blocked (readiness gate path) until loading/sync completes.
   - Also validates recovery after explicitly stripping the role label during the loading window, to guard against regressions of this fix.
2. e2e helper updates
   - Updated `e2e/util.go` helpers to use `ContainersReady` (instead of `PodReady`) where required for this intermediate-window check.

Abhra303:
@miledxz can you review?

miledxz:
hi @bwnuts thanks for the hard work ! 

please, there is one small required change before we merge, 

reconcileReplicaLabel can skip SLAVE OF based on stale metadata:

The label-only fast path here trusts checkReplicaRole, but that function will return true even when Dragonfly's live master_host points at a different master, as long as the pod's existing master-ip label/annotation still matches:

```
	if masterIp != redisMasterIp {
		if masterIp != pod.Labels[resources.MasterIpLabelKey] && masterIp != pod.Annotations[resources.MasterIpAnnotationKey] {
			return false, nil
		}
	}
```

So a pod that lost only its role label but kept a stale master-ip annotation could get re-labeled role=replica while still replicating from the wrong/old master — and we'd never issue SLAVE OF to correct it. The pre-branch flow always called configureReplica, so this is a behavior change specific to the new recovery path. 

suggestion, but feel free to use AI, and verify:

for the fast path, verify the live replication state instead of trusting pod metadata — i.e. only skip SLAVE OF when INFO replication reports role != master and master_host == sanitizeIp(masterIp). Otherwise fall through to configureReplica. checkReplicaRole can stay as-is for checkAndConfigureReplicas; only this recovery path needs the stricter check.

bwnuts:
@miledxz, thanks for the catch! I hope I got your point right.

what changed: 
`reconcileReplicaLabel` now uses a stricter live-state check for the label-only fast path:
- Added `checkReplicaLiveMaster(...)` and used it only in `reconcileReplicaLabel`.
- The fast path now applies **only if** `INFO replication` reports:
  - `role != master`, and
  - `master_host == sanitizeIp(masterIp)`.
- Otherwise it falls back to `configureReplica` (issues `SLAVE OF`) and updates metadata.

bwnuts:
@miledxz, please take a look

bwnuts:
@miledxz, could you please take a look, is there anything I need to add?

miledxz:
@bwnuts sorry for delay, I'm in process of reviewing it

PR #540, opened 2026-06-12
chore(manifests): update crd manifests

Abhra303:


miledxz (approved):
checked locally, lgtm

Issue #427, opened 2025-12-03
Helm chart: support for WATCH_NAMESPACE environment variable


menghanl:
## Feature request
Support for configuring the `WATCH_NAMESPACE` environment variable in the Helm chart.
Either by special case support for this one config, or as a general `extraEnv` value

## Background
The operator supports the WATCH_NAMESPACE environment variable, but there's currently no way to set it through the Helm chart values.

In one of our large clusters, the operator uses ~2GB memory when watching all namespaces. Setting WATCH_NAMESPACE to limit the watched namespaces reduces memory usage to ~13MB.

Abhra303:
@menghanl feel free to open a PR for this. Should be an easy addition.

PR #556, opened 2026-06-25
feat(chart): support for a namespaced scope deployment

cjabrantes:
## Summary

There are cases where its not allowed to deploy `ClusterRole` resources. 
This PR adds a `watchNamespaces` value to the Helm chart that enables a namespace-scoped deployment type — no `ClusterRole` required (when rbacProxy is not required).

The operator already supports the `WATCH_NAMESPACE` environment variable to restrict which namespaces it watches. When set, all resources it manages (`Dragonfly`, `StatefulSet`, `Pod`, `Service`, etc.) are namespace-scoped, so a `Role` per watched namespace is sufficient — no `ClusterRole` is needed for the manager.

This PR will also solve the https://github.com/dragonflydb/dragonfly-operator/issues/427.

## Changes

- **`values.yaml`**: new `watchNamespaces` value (empty by default — fully backward compatible)
- **`deployment.yaml`**: injects `WATCH_NAMESPACE` env var when `watchNamespaces` is set
- **`clusterroles.yaml`**: manager `ClusterRole` is skipped when `watchNamespaces` is set; `proxy-role`/`metrics-reader` are now correctly guarded by `rbacProxy.enabled` 
- **`clusterrolebindings.yaml`**: manager `ClusterRoleBinding` skipped when `watchNamespaces` is set; `proxy-clusterrolebinding` now correctly guarded by `rbacProxy.enabled` 
- **`roles.yaml`**: creates a namespace-scoped `Role` in each watched namespace when `watchNamespaces` is set
- **`rolebindings.yaml`**: creates a `RoleBinding` in each watched namespace when `watchNamespaces` is set

## Usage

To deploy without any `ClusterRole`:

```yaml
watchNamespaces: "my-namespace"   # comma-separated for multiple: "ns1,ns2"
rbacProxy.enabled: false
createAggregateRoles: false
```

Default behavior (empty watchNamespaces) is unchanged — the operator watches all namespaces and ClusterRole is used as before.

## Testing
Tested on a local cluster with 4 namespaces: operator, df1, df2, nodf (watchNamespaces: "df1,df2").

Operator deployed at operator
Dragonfly resource created in df1 and df2 — both reconciled correctly, no RBAC errors in logs
Dragonfly resource created in nodf — operator correctly ignored it 
Deleted the master pod in df1 — replica was promoted to master and old master reconfigured as replica
No ClusterRole resources deployed in the cluster

### DIFF
--- a/charts/dragonfly-operator/templates/clusterrolebindings.yaml
+++ b/charts/dragonfly-operator/templates/clusterrolebindings.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.watchNamespaces }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
@@ -13,8 +14,9 @@ subjects:
   - kind: ServiceAccount
     name: {{ include "dragonfly-operator.serviceAccountName" . }}
     namespace: {{ include "dragonfly-operator.namespace" . }}
+{{- end }}
+{{- if .Values.rbacProxy.enabled }}
 ---
-
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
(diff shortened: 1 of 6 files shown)


Abhra303 (approved):
Looks good, Thanks!

PR #559, opened 2026-07-08
chore(controller): configure takeover read timeout 

abhiramkaushik:
# Description

We have seen issues with the operator-initiated Dragonfly takeover flow. In one deployment with 1 master and 1 replica, `REPLTAKEOVER` can take longer than 40s to complete. The operator Redis client currently times out before the command finishes, which can lead to a retry of the takeover command. That retry can trigger `SIGABRT` in Dragonfly [here](https://github.com/dragonflydb/dragonfly/blob/main/src/server/server_family.cc#L3431).

This PR adds a configurable read timeout only for the `REPLTAKEOVER` path, so operators can set a larger timeout without changing other Redis client operations.

- Adds `--repl-takeover-read-timeout`, configurable via operator container args.
- Uses the configured timeout only for `REPLTAKEOVER`.
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -94,6 +94,11 @@ func main() {
 		os.Exit(0)
 	}
 
+	if err := validateReplTakeoverTimeout(replTakeoverTimeout); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 	mgrOpts := ctrl.Options{
@@ -258,4 +263,11 @@ func resolveOperatorNamespace(saFile string) string {
 		}
 	}
 	return ""
+}
+
+func validateReplTakeoverTimeout(timeout time.Duration) error {
+	if timeout <= 0 {
+		return fmt.Errorf("repl-takeover-read-timeout must be greater than 0")
+	}
+	return nil
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
@@ -71,9 +72,11 @@ func main() {
 	var versionFlag bool
 	var watchCurrentNamespace bool
 	var dragonflyImage string
+	var replTakeoverTimeout time.Duration
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&dragonflyImage, "dragonfly-image", "", "The default dragonfly image to use.")
+	flag.DurationVar(&replTakeoverTimeout, "repl-takeover-read-timeout", 10*time.Second, "The Redis client read timeout for REPLTAKEOVER.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
@@ -159,6 +162,7 @@ func main() {
 			EventRecorder:         eventRecorder,
 			DefaultDragonflyImage: dragonflyImage,
 			OperatorNamespace:     operatorNamespace,
+			ReplTakeoverTimeout:   replTakeoverTimeout,
 		},
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "Dragonfly")
@@ -172,6 +176,7 @@ func main() {
 			EventRecorder:         eventRecorder,
 			DefaultDragonflyImage: dragonflyImage,
 			OperatorNamespace:     operatorNamespace,
+			ReplTakeoverTimeout:   replTakeoverTimeout,
 		},
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "Health")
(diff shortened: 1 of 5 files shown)


cmd/main.go
```diff
@@ -71,9 +72,11 @@ func main() {
 	var versionFlag bool
 	var watchCurrentNamespace bool
 	var dragonflyImage string
+	var replTakeoverTimeout time.Duration
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&dragonflyImage, "dragonfly-image", "", "The default dragonfly image to use.")
+	flag.DurationVar(&replTakeoverTimeout, "repl-takeover-read-timeout", 10*time.Second, "The Redis client read timeout for REPLTAKEOVER.")
```
miledxz, line 79: REPLTAKEOVER is issued with a 10000 ms (10s) server-side timeout:

	resp, err := redisClient.Do(ctx, "repltakeover", "10000").Result()


The new flag defaults to the same 10s. When the client ReadTimeout equals the server command timeout, the client can still hit an i/o read timeout right as the server finishes the takeover.

Is closing this race the goal of the PR ? 

If so, the default should give some headroom over the command timeout (e.g. 15s, or command_timeout + buffer). At the default, the timeout behavior is effectively unchanged from before — users only benefit if they manually raise the flag.

internal/controller/dragonfly_instance.go
```diff
@@ -1119,7 +1133,8 @@ func (dfi *DragonflyInstance) updatedMaster(ctx context.Context, oldMaster *core
 func (dfi *DragonflyInstance) replTakeover(ctx context.Context, newMaster *corev1.Pod, oldMaster *corev1.Pod) error {
 	dfi.log.Info("running REPLTAKEOVER on replica", "pod", newMaster.Name)
 
-	redisClient := dfi.getRedisClient(newMaster.Status.PodIP)
+	redisClient := dfi.getReplTakeoverRedisClient(newMaster.Status.PodIP)
+	defer redisClient.Close()
 
 	resp, err := redisClient.Do(ctx, "repltakeover", "10000").Result()
```
miledxz: there is hidden coupling between --repl-takeover-read-timeout and the hardcoded 10000

command timeout here is a magic literal ("10000" = 10s), 

while the client read timeout is now independently configurable. Nothing ties the two together, so it's easy to misconfigure: e.g. --repl-takeover-read-timeout=2s makes the client give up after 2s while the server keeps working up to 10s — a guaranteed i/o read timeout even though the takeover may actually succeed server-side.  Which would be confusing failure mode.

The read timeout should always be >= the command timeout. A few options, roughly in order of preference:

- Make the command timeout a named constant (e.g. replTakeoverCommandTimeout) and derive the read timeout as commandTimeout + buffer, so they can't drift.
- Or keep the flag but validate readTimeout >= commandTimeout and reject/warn otherwise.

miledxz (changes requested):

cmd/main.go
abhiramkaushik: I think 10000 is in sec and not ms https://github.com/dragonflydb/dragonfly/blob/main/src/server/server_family.cc#L3407. 

Adding a few redacted log lines we saw in production:
```
---- master log ---------
2026-06-10T14:54:29.898028518Z INFO Takeover initiated, locking down the database.
-------------------------

---- replica log ---------
2026-06-10T15:03:41.998687069Z INFO Takeover successful, promoting this instance to master.
-------------------------
```
That was nearly 9 minutes of takeover (which we are still trying to figure out) and clearly the replica didn't time out when it initiated a take over [here](https://github.com/dragonflydb/dragonfly/blob/main/src/server/server_family.cc#L3437)

abhiramkaushik: @miledxz  would be able to verify
> I think 10000 is in sec and not ms https://github.com/dragonflydb/dragonfly/blob/main/src/server/server_family.cc#L3407.

please.

If this is the case, then there is a disconnect on what df expects vs what is sent.

abhiramkaushik:
@miledxz have you had the chance to look at this https://github.com/dragonflydb/dragonfly-operator/pull/559#discussion_r3553090377

internal/controller/dragonfly_instance.go
abhiramkaushik: > --repl-takeover-read-timeout=2s makes the client give up after 2s while the server keeps working up to 10s

yes, this is exactly whats happening right now as well.

> Make the command timeout a named constant (e.g. replTakeoverCommandTimeout) and derive the read timeout as commandTimeout + buffer, so they can't drift.

This adds a hidden buffer which can be confusing to someone.

> Or keep the flag but validate readTimeout >= commandTimeout and reject/warn otherwise.

this is better right? its visible, and errors immediately indicating that the configuration requires a fix

cmd/main.go
abhiramkaushik: tested this out locally and updated to include a buffer of 5seconds for the client and the takeover itself is equal to the amount of read time out

ashotland:
@abhiramkaushik is intentionally short as master is not serving during the takeover 

dragonfly should not crash when takeover is retried.
can you share more evidence on the crash? what dragonfly version did you observe it in?

abhiramkaushik:
> @abhiramkaushik is intentionally short as master is not serving during the takeover
> 
> dragonfly should not crash when takeover is retried. can you share more evidence on the crash? what dragonfly version did you observe it in?

@ashotland dragonfly 1.39+ has had a few patches which fixes the problems is dragonfly and I think this PR is no longer necessary. 


> @abhiramkaushik is intentionally short as master is not serving during the takeover

Look at this
 >I think 10000 is in sec and not ms https://github.com/dragonflydb/dragonfly/blob/main/src/server/server_family.cc#L3407.

The main issue in the operator is that it sends "10000" as the timeout parameter which on the dragonfly side is used as seconds and not milliseconds. This part would need to be fixed, unless it has already been fixed in the latest operator version

ashotland:
Thanks @abhiramkaushik - you are correct repltakeover timeout is in seconds we should lower it.

abhiramkaushik:
> Thanks @abhiramkaushik - you are correct repltakeover timeout is in seconds we should lower it.

Yes, thanks! i can raise a PR for it if needed, but if you've already got it, then i'd be happy to just wait

PR #560, opened 2026-07-09
revert(operator): remove operator-managed NetworkPolicy

### LINKED ISSUE #495: NetworkPolicy blocks metrics scraping from external pods
## Summary

The operator-generated `NetworkPolicy` for a `Dragonfly` instance restricts ingress on the admin/metrics port (9999) to only the operator's own controller-manager and peer Dragonfly pods. This makes it impossible for any external scraper (Prometheus, VictoriaMetrics, etc.) to reach the `/metrics` endpoint, even though exposing metrics is an explicitly supported feature of the admin port.

## Environment

- DragonflyDB Operator (1.4.0 and 1.5.0)
- k3s with Flannel CNI
- VictoriaMetrics (vmagent) as scraper, configured via PodMonitor

## Steps to Reproduce

1. Deploy a `Dragonfly` instance via the operator
2. Configure a `PodMonitor` or `ServiceMonitor` targeting port 9999 (`admin`)
3. Observe scrape errors: `dial tcp <pod-ip>:9999: connect: connection refused`

## Root Cause

The operator creates a `NetworkPolicy` with the following ingress rules:

```yaml
ingress:
  - ports:
      - port: 6379
        protocol: TCP
  - from:
      - namespaceSelector: {}
        podSelector:
          matchLabels:
            control-plane: controller-manager
      - podSelector:
          matchLabels:
            app: app-dragonfly
            app.kubernetes.io/name: dragonfly
            app.kubernetes.io/part-of: dragonfly
    ports:
      - port: 9999
        protocol: TCP
```

Port 6379 (Redis) is open to all pods, but port 9999 (admin/metrics) is scoped exclusively to the operator controller and peer Dragonfly pods. Any scraper pod — regardless of namespace — is refused at the network level before it can reach the listener.

This is confirmed by:
- `/proc/net/tcp` showing the admin port correctly bound on `0.0.0.0:9999`
- `kubectl port-forward` successfully reaching `/metrics` (bypasses pod networking)
- Cross-pod `curl` to the pod IP returning immediate `connection refused` (RST, not timeout — i.e. actively blocked, not dropped)
- Removing/patching the NetworkPolicy immediately resolves scraping

## Expected Behavior

The NetworkPolicy should permit ingress on port 9999 from any pod that can reach port 6379, or at minimum provide a documented mechanism (e.g. a field on the `Dragonfly` CR) to allow scraper pods access to the admin port.

## Workaround

Manually add an additional ingress rule to the NetworkPolicy after deployment. Note that the operator may reconcile this away:

```yaml
- ports:
    - port: 9999
      protocol: TCP
```

Or as a separate `NetworkPolicy` object (which the operator will not overwrite):

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: dragonfly-allow-metrics
  namespace: <your-namespace>
spec:
  podSelector:
    matchLabels:
      app.kubernetes.io/name: dragonfly
  ingress:
    - ports:
        - port: 9999
          protocol: TCP
```

## Suggested Fix

Add a `allowedMetricsSources` (or similar) field to the `Dragonfly` CRD so users can specify a `podSelector`/`namespaceSelector` for their scraper, which the operator then includes in the generated NetworkPolicy.

Example in MetalLB: https://github.com/metallb/metallb/blob/7f8c364490486db29c2f588d466acf07ee06861f/charts/metallb/values.yaml#L54

### LINKED ISSUE #546: Cluster access restriction in v1.6.0 should be advertised better
The introduction of cluster access (port 6379) restriction in v1.6.0 (#514), is very disruptive for users not running database and workload in the same namespace. I think most k8s clusters are setup with databases in separate namespaces, so this probably hit most users.
I'm all for plugging security holes, but I would really appreciate if changes like this would be better advertised in release notes going forward. This would make it so much easier to ensure a smooth upgrade process for all users.

I've two proposals for the future:
1) Add explicit callouts to disruptions like this one in the release notes. This was a breaking change for most users!
2) If possible template all `Networkplolicies` in the helm chart. This will ensure that CI workflows will pick the changes, and make it more clear to the end users.

I'll leave a `NetworkPolicy` here that'll open up access to a cluster from other namespaces again. Hopefully this will help others having the same issue as I had.

```yaml
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: dragonfly-cluster-access
spec:
  podSelector:
    matchLabels:
      app: dragonfly # <- dragonfly cluster name
      app.kubernetes.io/name: dragonfly
      app.kubernetes.io/part-of: dragonfly
  policyTypes:
    - Ingress
  ingress:
    - from:
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: <externalNamespace> # <- namespace with workload
          podSelector:
            matchLabels:
              app.kubernetes.io/name: <app> # <- restrict to single app if needed
      ports:
        - protocol: TCP
          port: 6379
```

miledxz:
Removes the operator-managed `NetworkPolicy` feature that was introduced in #473 and hardened in #514. 

The generated policy restricted:

- **client port `6379`** to the same namespace only
- **admin port `9999`** to the operator and peer Dragonfly pods only

This was disruptive for the common setup where databases and workloads live in separate namespaces (#546) and blocked external Prometheus / VictoriaMetrics scrapers from reaching `/metrics` (#495).

This PR fully reverts the feature to the pre-#473 state. Users who need network restrictions can apply their own `NetworkPolicy`, which the operator will no longer reconcile away.

## Breaking change

Deployments relying on the operator to create a `NetworkPolicy` (via the default or `networkPolicyEnabled: true`) will no longer get one. The `networkPolicyEnabled` field is removed from the CRD; any manifests setting it must drop the field.

Fixes #546
Fixes #495


moredure (approved):

ashotland:
Should we add some tests to avoid re-introducing that ?

miledxz:
> Should we add some tests to avoid re-introducing that ?

yes, we can do it for sure

charts/dragonfly-operator/templates/deployment.yaml
```diff
@@ -67,8 +67,6 @@ spec:
             {{- toYaml .Values.rbacProxy.resources | nindent 12 }}
           {{- end }}
         - name: manager
-          command:
```
moredure, line 70: version of chart should be updated as well?

moredure:
@miledxz is it worth splitting into changes to the app and changes to the helm chart?

miledxz:
> @miledxz is it worth splitting into changes to the app and changes to the helm chart?

I think we can do that

charts/dragonfly-operator/templates/deployment.yaml
miledxz: I think yes but not sure cc @Abhra303

Abhra303: right now we release both with same version at the same time.

Abhra303:
> @miledxz is it worth splitting into changes to the app and changes to the helm chart?

yes, it is worth it to split.

alx8:
**Production data point supporting this revert — v1.6.1, entire ArgoCD installation down**

Adding a concrete incident in case it helps prioritise merging, since this has been approved for a while.

On 2026-08-05 we upgraded dragonfly-operator to v1.6.1 on a production cluster and it took our whole ArgoCD installation down.

The part that makes this sharper than #546: in v1.6.1 the CRD ships `networkPolicyEnabled` with `default: true`. Users who never opted in, and whose manifests don't mention the field, get the NetworkPolicy created for them on upgrade. There is no opt-out you can pre-set beforehand if you don't know the field exists — and it isn't visible in a values diff, because the CRD is templated rather than living in `crds/`.

Generated policy on our cluster:

```yaml
spec:
  podSelector:
    matchLabels:
      app: dragonfly-ops
      app.kubernetes.io/name: dragonfly
      app.kubernetes.io/part-of: dragonfly
  policyTypes: [Ingress]
  ingress:
  - from:
    - podSelector: {}          # same namespace only
    ports:
    - {port: 6379, protocol: TCP}
```

**Impact.** ArgoCD uses Dragonfly as its Redis cache from another namespace. All 29 Applications went `Sync: Unknown` and auto-sync refused on every one of them — status is Unknown *because* the resource tree can't be written to the cache. **ArgoCD therefore could not deploy its own fix**; recovery required a manual `kubectl patch` on the Dragonfly CRs. A separate app also lost its session store and began failing readiness probes.

**Two things worth noting for anyone diagnosing this:**

1. NetworkPolicy drops are silent, so clients report `i/o timeout`, not `connection refused`. That distinction is the fastest way to tell a policy block from a dead listener.
2. Dragonfly itself stays perfectly healthy throughout — pod Running, CR `Ready`, Service endpoints present. Nothing about the Dragonfly instance points at the cause.

**One request on the removal itself.** Since this PR drops `networkPolicyEnabled` from the CRD, please call out in the release notes that manifests setting it must remove the field. That's the same advertising gap #546 asked for, just in the other direction — anyone who worked around the v1.6.x default by setting `networkPolicyEnabled: false` will be carrying that field when the revert lands.

Separately, while this remains unmerged: the CRD field's description currently reads *"The NetworkPolicy restricts admin port access to operator and peer pods only."* That describes only the 9999 rule, not the 6379 one that actually causes the breakage. If this PR stalls further, correcting that description would be a cheap standalone improvement.

Abhra303:
@miledxz can you fix the conflict and merge it please?

miledxz:
@copilot resolve the merge conflicts in this pull request

miledxz:
@Abhra303 I will test one more time in kind and we can merge after that

Issue #238, opened 2024-08-27
Better handling of master failover


laurivosandi:
It looks like most applications do have Redis reconnect support if Redis operation results in error such as `ReplyError: READONLY You can't write against a read only replica.`

This is related to https://github.com/dragonflydb/dragonfly-operator/issues/149#issuecomment-1921063199

To better support dumb clients I would advise implementing TCP disconnect when Dragonfly pod loses master role to ensure clients reconnect to the new Dragonfly master pod

Pothulapati:
Hmm, This makes sense but

> To better support dumb clients I would advise implementing TCP disconnect when Dragonfly pod loses master role to ensure clients reconnect to the new Dragonfly master pod

How can we do this? Considering we don't have a proxy to do different replies. It's upto the older pod to return READONLY (which they will once they are configured as a replica). wdyt? 🤔

nmass-betpawa:
Any updates regarding this problem?

Pothulapati:
> I think it would be fine to issue TCP connection reset for every open connection if pod loses write master status. You can make this behaviour optional via command line flag or something like that

Functionality to kill client connections currently does not exist in Dragonfly (See https://github.com/dragonflydb/dragonfly/issues/2178). and there is no common Kubernetes abstraciton/functionality to do this. :/

cc: @romange

jonathon2nd:
Hello!

I we are trying to do some tests with failover and wondering the best way to do this. Don't want to delete the master pod, as that always works. 

Want to issue a failover, but 
```
127.0.0.1:6379> failover
(error) ERR unknown command `FAILOVER`
```

And when I try manually
```
127.0.0.1:6379> REPLICAOF NO ONE
OK
```
on replica works, but then when I try to get the old master to follow new one this happens. 
```
127.0.0.1:6379> REPLICAOF 10.42.90.34 6379
(error) ERR could not greet master Bad message
```

Any ideas?

jonathon2nd:
found reference to a command called REPLTAKEOVER, but does not seem like I can get it to do anything, help is also empty
```
10.43.107.71:6379> REPLTAKEOVER "10.42.206.142" "6379"
(error) ERR value is not an integer or out of range
10.43.107.71:6379> REPLTAKEOVER
(error) ERR wrong number of arguments for 'repltakeover' command
10.43.107.71:6379> help REPLTAKEOVER

```

alphabet5:
[...]

PR #575, opened 2026-08-14
fix(controller): don't lose the client disconnect when it fails

### LINKED ISSUE #324: Drop connected clients when instance is no longer master
Previously reported in https://github.com/dragonflydb/dragonfly/issues/5160

**Describe the bug**
We encountered a situation where Dragonfly decided to switch masters (without any real reason as far as I can tell, but that's for another topic). We are using the operator in kubernetes. Our apps connect to dragonfly via the service, so Kubernetes routes the traffic to the current master.

When the master changes, open connections don't magically re-route to the new master but instead stay connected to the now no-longer-master. Because this instance is now read-only, writing to it will fail:

```
READONLY You can't write against a read only replica.
```

Most of the times a failover will occur because the old master is no longer alive, which already kills open connections. But in a "planned switchover" (or whatever Dragonfly calls this internally), the old master remains intact and connected.

**To Reproduce**
1. Deploy the operator with 2 or more replicas
2. Have any app be connected via the service that tries to write every second.
3. Trigger a switchover without killing the present master

**Expected behavior**
If dragonfly does a role change, it should give connected clients some indication that they need to reconnect in order to talk to the new master. A simple disconnect would be very effective.

As per suggestion of @romange:
> The operator can send "client kill" command

**Environment (please complete the following information):**
 - Containerized?: Kubernetes via Operator
 - Dragonfly Version: 1.30.0

**Logs from master replica**
```
I20250520 12:36:00.022531    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
I20250520 12:37:00.020418    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
I20250520 12:38:00.019855    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
I20250520 12:39:00.018941    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
I20250520 12:39:08.867802    11 dflycmd.cc:127] Disconnecting from replica 10.244.0.57:6379
W20250520 12:39:08.867852    11 common.cc:413] ReportError: Operation canceled: ExecutionState cancelled
I20250520 12:39:08.867894    11 dflycmd.cc:686] Replication error: Operation canceled: ExecutionState cancelled
I20250520 12:39:09.960482    11 server_family.cc:3009] Replicating 10.244.0.57:9999
I20250520 12:39:09.964658    11 replica.cc:580] Started full sync with 10.244.0.57:9999
I20250520 12:39:09.989066    11 replica.cc:600] full sync finished in 26 ms
I20250520 12:39:09.989102    11 replica.cc:690] Transitioned into stable sync
I20250520 12:40:00.019615    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
I20250520 12:41:00.019866    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
I20250520 12:42:00.017843    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
I20250520 12:43:00.018100    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
W20250520 12:43:28.733081    11 script_mgr.cc:328] Error running script (call to a08c9c69b7c07ae2485190873b90d128a23e502d): @user_script:2: -READONLY You can't write against a read only replica.
W20250520 12:43:28.733228    11 main_service.cc:1363]  EVAL return redis.call(\'exists\',KEYS[1])<1 and redis.call(\'setex\',KEYS[1],ARGV[2],ARGV[1]) 1 ";N;} 900 failed with reason: Error running script (call to a08c9c69b7c07ae2485190873b90d128a23e502d): @user_script:2: -READONLY You can't write against a read only replica.
W20250520 12:43:36.952080    11 script_mgr.cc:328] Error running script (call to a08c9c69b7c07ae2485190873b90d128a23e502d): @user_script:2: -READONLY You can't write against a read only replica.
W20250520 12:43:36.952189    11 main_service.cc:1363]  EVAL return redis.call(\'exists\',KEYS[1])<1 and redis.call(\'setex\',KEYS[1],ARGV[2],ARGV[1]) 1 ";N;} 900 failed with reason: Error running script (call to a08c9c69b7c07ae2485190873b90d128a23e502d): @user_script:2: -READONLY You can't write against a read only replica.
I20250520 12:44:00.018446    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
I20250520 12:45:00.018730    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
I20250520 12:46:00.019562    11 save_stages_controller.cc:342] Saving "/dragonfly/snapshots/snapshot-summary.dfs" finished after 1 s
```

krangerich:
During a failover, the dragonfly operator sends a `SLAVE OF <new master>` to the old master, sets the label to `role: replica`  and then trys to disconnect clients. But disconnecting clients isn't robust and might fail. 

A timeout on `CLIENT LIST` left the old master's clients connected for good: the error was only logged, and the next reconcile finds a correctly configured replica, so replicaOf never runs for that pod again.

Mark the pod before demoting it and clear the mark once the disconnect went through. Disconnecting the clients is now retried by a reconciliation loop. 
 `CLIENT LIST` plus one kill per client becomes a single `CLIENT KILL LADDR`. `CLIENT KILL LADDR` was added in Dragonfly v1.14.0, together with the previously used  `CLIENT KILL`. 

Fixes #324
Related: #238

### DIFF
--- a/internal/controller/dragonfly_instance.go
+++ b/internal/controller/dragonfly_instance.go
@@ -478,6 +478,12 @@ func (dfi *DragonflyInstance) replicaOf(ctx context.Context, pod *corev1.Pod, ma
 		return fmt.Errorf("failed to determine the current role of the instance: %w", err)
 	}
 
+	if wasMaster {
+		if err := dfi.setPendingClientDisconnect(ctx, pod, true); err != nil {
+			return fmt.Errorf("could not mark pod for client disconnect: %w", err)
+		}
+	}
+
 	// Sanitize masterIp in case ipv6
 	masterIp = sanitizeIp(masterIp)
 
@@ -518,7 +524,16 @@ func (dfi *DragonflyInstance) replicaOf(ctx context.Context, pod *corev1.Pod, ma
 
 	if wasMaster {
 		// Prevent clients from sending commands to this old master
-		dfi.disconnectClients(ctx, redisClient, pod)
+		if err := dfi.disconnectClients(ctx, pod); err != nil {
+			dfi.log.Error(err, "failed to disconnect clients, will retry", "pod", pod.Name)
+			// Returning an error here would abort the caller's loop over the remaining pods.
+			// The annotation stays set, the next reconcile retries.
+			return nil
+		}
+
+		if err := dfi.setPendingClientDisconnect(ctx, pod, false); err != nil {
+			return fmt.Errorf("could not clear the client disconnect marker: %w", err)
+		}
 	}
 
 	return nil
@@ -557,6 +572,7 @@ func (dfi *DragonflyInstance) replicaOfNoOne(ctx context.Context, pod *corev1.Po
 		pod.Annotations = make(map[string]string)
 	}
 	pod.Annotations[resources.MasterIpAnnotationKey] = masterIp
+	delete(pod.Annotations, resources.PendingClientDisconnectAnnotationKey)
 
 	if err := dfi.client.Patch(ctx, pod, patch); err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


dimoschi:
Thanks for this. The lost-disconnect bug is real, I hit it too: a timed out `CLIENT LIST` clears the obligation and the pod is never revisited, so the clients stay attached. `CLIENT KILL LADDR` in one call is also better than the list-and-kill loop.

I stacked a follow-up on your branch as #583. I had written my own marker before I saw your PR and picked the same annotation key, character for character, so yours should be the one that lands. #583 keeps your commit and adds one on top.

**I do not think #575 alone closes #324.** The retry only runs when the disconnect fails. In the normal path `CLIENT KILL` succeeds first time, the marker clears, and we are done. But the kill still happens right after the role label patch, while the pod is still in the master Service's endpoint slices. A client that reconnects in that window goes back to the pod being demoted, gets the read-only instance again, and nothing retries, because the disconnect is already marked done.

```
t0  operator patches role=replica  -> API server accepts
t1  EndpointSlice controller updates the slice
t2  kube-proxy / mesh proxies apply it
t3  new connections stop reaching the demoted pod
```

#436 named this condition in its commit message, "after it's removed from the service", but never waited for it.

Measured with php-fpm and phpredis persistent connections, 300 requests, failover at request 80:

| kill timing | ok | failed | recovers |
|---|---|---|---|
| none | 79 | 221 | never |
| right after demotion | 79 | 221 | never |
| after the Service converged | 300 | 0 | yes |

Rows 1 and 2 are identical. At t0 the kill buys nothing. `TestReplicaOfDoesNotDisconnectClientsWhileStillInTheMasterService` in #583 fails against your branch and passes with the commit on top, so you can check this rather than trust it.

One thing to catch before merge: **both PRs say `Fixes #324`.** If yours merges first, the issue closes while the bug still reproduces. Drop it from one of them.

Fine either way on shape: land yours and #583 follows, or fold the wait into your PR and close mine.

PR #578, opened 2026-08-23
feat(operator): auto-create Dragonfly credentials secret

LCerebo:
Opt-in functionality to generate a secret with credentials and connection details for the cluster, if the secret is deleted/update the cluster will be updated too.
The functionality was inspired by the cnpg operator.
This cover the request of @laurivosandi in https://github.com/dragonflydb/dragonfly-operator/issues/50#issuecomment-1598669833

### DIFF
--- a/api/v1alpha1/dragonfly_types.go
+++ b/api/v1alpha1/dragonfly_types.go
@@ -305,6 +305,14 @@ type Snapshot struct {
 }
 
 type Authentication struct {
+	// (Optional) When enabled, the operator auto-creates the password Secret.
+	// If passwordFromSecret is omitted, the Secret name defaults to <cluster-name>-password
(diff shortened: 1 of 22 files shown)


ashotland:
IMO it is not the operator's responsibility to create and manage secrets

LCerebo:
> IMO it is not the operator's responsibility to create and manage secrets

When external vault integration or credentials policy like secret rotation are required the `passwordFromSecret`
 method is the preferred one. This is why the feature is opt-in, when enabled it also ease the connection to the cluster from clients that can access the auto generated secret. 
With this feature one can deploy a cluster with bootstrapped authentication using only a single resource and the operator guarantees that the secret and the password used by the cluster stays in sync.

